PR #46, opened 2017-02-16
Warn about browser plugins if PeerConnection fails

dbrgn:
@sillych adding these warnings might make troubleshooting easier for some people.


rugk:
Hmm, but 'ad blocker' is the wrong term. I only know uBlock, which is an ad blocker, but also describes itself as a general purpose-blocker, which does this.
Most plugins doing this are not ad blockers,…

At least in the German version the "whcih also blocks WebRTC is missing".

PR #49, opened 2017-02-16
Add referrer policy to HTML

rugk:
You already use a HTTP header on your web site, but users hosting their own-version of Threema Web or using a local HTML version may not get this header, so better specify it in the HTML source (too).


dbrgn:
Thanks!

I'm not sure why the tests failed, probably a configuration issue (I'm new to CircleCI). Could you rebase on the current master and force-push again to see if the tests run through this time?

PR #67, opened 2017-02-20
Show warning if browser blocks local storage

dbrgn:
This can happen through add-ons or when blocking website content. Here e.g. in Chrome:

chromesetting

This is the warning we show:

localstorage-de
localstorage-en


dbrgn:
@sillych good catch. Better now?

PR #77, opened 2017-02-21
Autofocus compose area on opening a conversation

### LINKED ISSUE #41: Message input edit line is not automatically focused when clicking on a contact
When clicking on a contact in Threema Web, the message input edit line is not automatically focused (like it is in WhatsApp Web), thus forcing the user to move the mouse to the input edit line and clicking into it (which sucks usability-wise).
The expected behaviour is to be able to start typing right away after clicking on a contact.

sillych:
Fixes #41


wschlich:
When will this go live on web.threema.ch? :-)

PR #80, opened 2017-02-21
Append message id to downloading files 

### LINKED ISSUE #61: webclient save images all under the same name (keep original filename)
If I try so save multiple images from my browser (Chrome) every file is called image.jpg, very annoying to manually rename them. Keeping the original filename intact would be really handy, also in telling the person that send the photos about which photo I am talking.

sillych:
Fixes #61

### DIFF
--- a/src/services/message.ts
+++ b/src/services/message.ts
@@ -105,6 +105,48 @@ export class MessageService implements threema.MessageService {
     }
 
     /**
+     * return the filename of a message (image, audio, file)
+     * used for downloads
+     * @param message
+     * @returns filename string or null
+     */
+    public getFileName(message: threema.Message): string {
+        if (message === undefined) {
+            return null;
+        }
+
+        let getFileName = (prefix: string, postfix: string = null): string => {
+            return prefix
+                + '-' + message.id
+                + (postfix !== null ? '.' + postfix : '');
+        };
+
+        switch (message.type) {
+            case 'image':
+                return getFileName('image', 'jpg');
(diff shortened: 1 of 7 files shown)


src/services/message.ts
```diff
@@ -105,6 +105,48 @@ export class MessageService implements threema.MessageService {
...
+     */
+    public getFileName(message: threema.Message): string {
+        if (message === undefined) {
+            return null;
+        }
+
+        let getFileName = (prefix: string, postfix: string = null): string => {
+            return prefix
+                + '-' + message.id
+                + (postfix !== null ? '.' + postfix : '');
```
dbrgn: Instead of using an explicit `null` default value, you could also make the postfix optional, in which case it'll be `undefined`:

```typescript
let getFileName = (prefix: string, postfix?: string): string => {
    // ...
    + (postfix !== undefined ? '.' + postfix : '');
};
```

src/services/message.ts
```diff
@@ -105,6 +105,48 @@ export class MessageService implements threema.MessageService {
...
+            case 'file':
+                if (message.file.type === 'image/gif') {
+                    return getFileName('animgif', 'gif');
+                }
+
+                if (message.file !== undefined) {
+                    return message.file.name;
+                }
+
+                // should not happen
```
dbrgn: Maybe log a warning?

sillych: good catch, done!

sillych:
done

PR #83, opened 2017-02-22
ICE Connectivity

lgrahl:
This PR should resolve most of the problems described in  #43 by enabling the use of TURN-TCP. I assume the problem was that the STUN/TURN requests have used UDP over port 443 which was blocked by the (corporate) firewalls.

Any remaining connectivity issues should be fixable by enabling TURN-over-TLS but this requires some modification to the TURN server configuration.

I also added some ICE candidate logging.

### DIFF
--- a/src/config.ts
+++ b/src/config.ts
@@ -16,14 +16,11 @@ export default {
     SALTYRTC_SERVER_KEY: 'b1337fc8402f7db8ea639e05ed05d65463e24809792f91eca29e88101b4a2171',
 
     // ICE
-    SALTYRTC_STUN: {
-        urls: 'stun:stun.threema.ch:443',
-    },
-    SALTYRTC_TURN: {
+    ICE_SERVERS: [{
         urls: 'turn:turn.threema.ch:443',
         username: 'threema-angular',
         credential: 'Uv0LcCq3kyx6EiRwQW5jVigkhzbp70CjN2CJqzmRxG3UGIdJHSJV6tpo7Gj7YnGB',
-    },
+    }],
 
     // Push
     PUSH_URL: 'https://push-web.threema.ch/push',
--- a/src/config.ts
+++ b/src/config.ts
@@ -17,7 +17,10 @@ export default {
 
     // ICE
     ICE_SERVERS: [{
-        urls: ['turn:turn.threema.ch:443', 'turn:turn.threema.ch:443?transport=tcp'],
+        urls: [
+            'turn:turn.threema.ch:443?transport=udp',
+            'turn:turn.threema.ch:443?transport=tcp'
+        ],
         username: 'threema-angular',
         credential: 'Uv0LcCq3kyx6EiRwQW5jVigkhzbp70CjN2CJqzmRxG3UGIdJHSJV6tpo7Gj7YnGB',
     }],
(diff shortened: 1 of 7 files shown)


dbrgn:
Awesome, thanks @lgrahl! I always assumed that TURN would try both UDP and TCP by default when not specifying any transport.

This will probably resolve the issues for all firewalls that don't use deep packet inspection.

lgrahl:
I would have assumed that as well. I'll put this on my to do list for investigation.

src/config.ts
```diff
@@ -16,14 +16,11 @@ export default {
     SALTYRTC_SERVER_KEY: 'b1337fc8402f7db8ea639e05ed05d65463e24809792f91eca29e88101b4a2171',
 
     // ICE
-    SALTYRTC_STUN: {
-        urls: 'stun:stun.threema.ch:443',
-    },
-    SALTYRTC_TURN: {
-        urls: 'turn:turn.threema.ch:443',
+    ICE_SERVERS: [{
+        urls: ['turn:turn.threema.ch:443', 'turn:turn.threema.ch:443?transport=tcp'],
```
dbrgn: Do you think the first URL should be explicitly marked with `?transport=udp`?

[RFC7065](https://tools.ietf.org/html/rfc7065) does not specify what protocol to use when not specifying a transport explicitly. So it must be up to the browser.

lgrahl: I suppose it doesn't hurt, yeah.

src/services/peerconnection.ts
```diff
@@ -135,6 +138,12 @@ export class PeerConnectionHelper {
         };
         this.webrtcTask.on('candidates', (e: saltyrtc.tasks.webrtc.CandidatesEvent) => {
             for (let candidateInit of e.data) {
+                if (candidateInit) {
+                    this.$log.debug(this.logTag, 'Adding remote ICE candidate:',
+                        PeerConnectionHelper.censorCandidate(candidateInit.candidate));
+                } else {
+                    this.$log.debug(this.logTag, 'No more remote ICE candidates');
```
dbrgn: So does this mean that in some cases the `e.data` attribute can contain a `null` value? I never experienced that in my tests.

lgrahl: Huh, you're right, we haven't specified that in SaltyRTC's WebRTC task. However, WebRTC does use `null` to indicate end of candidates (as can be seen in line 98) and we actually should pass that information to the stack to possibly improve the gathering process.

By the way, it has been specified for the [SaltyRTC ORTC Task](https://github.com/saltyrtc/saltyrtc-meta/blob/master/Task-ORTC.md#ice-candidates-message).

lgrahl:
I'm done. As @dbrgn mentioned, we probably need to update the SaltyRTC WebRTC Task Spec to (possibly) speed up the connectivity process but that is independent from this PR.

lgrahl:
Probably don't have to tell you that this needs to be done on the app side as well, right? :)

lgrahl:
I also did some tests whether `?transport=tcp` is sufficient that UDP relay candidates are being allocated. Short answer: No, we need both.

dbrgn:
Thanks! Merged it manually at 787396daf6aaff6b22ffabece7e52881efd1555a.

PR #84, opened 2017-02-22
Add support for pasting data into compose area 

### LINKED ISSUE #82: Support pasting image data
## Expected Behavior

When copying image or gif data through the clipboard and pasting it into the compose area, the image should be sent.

This will *revolutionize* sending of animgifs!

## Current Behavior

Nothing happens.

## Possible Solution

https://stackoverflow.com/questions/490908/paste-an-image-from-clipboard-using-javascript

dbrgn:
Fixes #82.

### DIFF
--- a/src/directives/compose_area.ts
+++ b/src/directives/compose_area.ts
@@ -283,7 +283,7 @@ export default [
                             let fileName: string;
                             if ((blob as any).name) {
                                 fileName = (blob as any).name;
-                            } else if (blob.type && blob.type.indexOf('/') !== -1) {
+                            } else if (blob.type && blob.type.match(/^[^;]*\//) !== null) {
                                 const fileExt = blob.type.split(';')[0].split('/')[1];
                                 fileName = 'clipboard.' + fileExt;
                             } else {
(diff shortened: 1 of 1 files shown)


src/directives/compose_area.ts
```diff
@@ -239,22 +246,78 @@ export default [
...
+                        const reader = new FileReader();
+                        reader.onload = function() {
+                            let buffer: ArrayBuffer = this.result;
+
+                            // Construct file name
+                            let fileName: string;
+                            if ((blob as any).name) {
+                                fileName = (blob as any).name;
+                            } else {
+                                const fileExt = blob.type.split(';')[0].split('/')[1];
```
sillych: What if type is string does not contain a `/`?

dbrgn:
@sillych better now?

src/directives/compose_area.ts
```diff
@@ -283,7 +283,7 @@ export default [
                             let fileName: string;
                             if ((blob as any).name) {
                                 fileName = (blob as any).name;
-                            } else if (blob.type && blob.type.indexOf('/') !== -1) {
+                            } else if (blob.type && blob.type.match(/^[^;]*\//) !== null) {
```
sillych: regex :trumpet:

PR #96, opened 2017-02-25
Fix MIME type icon path

rugk:
This was not subdir-compatible…


dbrgn:
Thanks. Your commit is based on a old version of Threema Web, therefore the tests don't run.

Could you please rebase your changes on the current version of Threema Web? In case the upstream remote is called "upstream", that would be:

    git checkout master
    git pull upstream master
    git checkout fix-mime-path
    git rebase master

PR #98, opened 2017-02-27
Replace uses of iterators on NodeList elements with for loop

### LINKED ISSUE #64: el[0].childNodes[Symbol.iterator] is not a function
After connecting my samsung s4 (sgh-m919) android (4.4.4) to my chrome (Version 46.0.2490.71 (64-bit)) linux (mint 17) pc - it all shows, and I click a user to send a message to - and it populates my browser with recent messages.  I type a new message to send in the box at the bottom, but when I click the split triangle on the right (hover state even changes) - nothing happens - and the text remains in the box.

dbrgn:
Apparently some older browser versions don't support for-of loops on `NodeList` elements yet: https://developer.mozilla.org/en/docs/Web/API/NodeList

Replace all occurrences of for-of loops on `.childNodes` with explicit for loops.

Fixes #64.

### DIFF
--- a/src/directives/compose_area.ts
+++ b/src/directives/compose_area.ts
@@ -115,7 +115,8 @@ export default [
                 // Emoji images are converted to their alt text in this process.
                 function submitText(): Promise<any> {
                     let text = '';
-                    for (let node of composeDiv[0].childNodes) {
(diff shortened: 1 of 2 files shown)


dbrgn:
Fixes #64.

PR #103, opened 2017-02-27
Escape pasted text in compose area

### LINKED ISSUE #86: Pasting text with '<' and '>' doesn't work correctly
## Expected Behavior

When pasting text like 
`lorem ipsum x<y blah blah` or `lorem ipsum x<y blah blah y>z foo bar` into the message text field, exactly this text should appear in the text field.

## Current Behavior

The first text gets cut off at the '<' character, only `lorem ipsum x` appears in the text field.
In the second sample, everything between the two brackets disappears, resulting in `lorem ipsum xz foo bar`.

## Possible Solution

There is probably something up with [the regexes in filter.ts](https://github.com/threema-ch/threema-web/blob/master/src/filters.ts#L64-L65), I couldn't find anything obvious, though.

1. Copy `lorem ipsum x<y blah blah y>z foo bar` into your clipboard
2. Paste into the message text field
3. Compare the text in the text field with the text in your clipboard

## Your Environment

* Threema Web version: 1.0.2
* Threema for Android version: 3.01
* Browser name and version: Chrome 56
* Operating system and version: macOS 10.12

dbrgn:
Sanitizing text would cause some non-HTML text to disappear (see #86). Instead, we escape HTML, so that it looks exactly like pasted.

The applyFilters helper function has been removed, it's unclear what it does. Being explicit is better in this case.

Fixes #86.

### DIFF
--- a/src/directives/compose_area.ts
+++ b/src/directives/compose_area.ts
@@ -26,7 +26,6 @@ export default [
     '$translate',
     '$mdDialog',
     '$filter',
-    '$sanitize',
     '$log',
     function(browserService: threema.BrowserService,
              stringService: threema.StringService,
@@ -34,7 +33,6 @@ export default [
              $translate: ng.translate.ITranslateService,
              $mdDialog: ng.material.IDialogService,
              $filter: ng.IFilterService,
-             $sanitize: ng.sanitize.ISanitizeService,
              $log: ng.ILogService) {
         return {
             restrict: 'EA',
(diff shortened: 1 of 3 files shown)


dbrgn:
@sillych I'm not sure whether we should remove the htmlToAsciiMarkup filter. What do you think happens more often, people pasting rich formatted text (e.g. bold) from a website or an e-mail and wanting it to be formatted, or people sharing HTML markup with coworkers? In the latter case, they'd want the HTML code to remain unchanged.

PR #110, opened 2017-02-28
Prototype for Settings & Darktheme

FloThinksPi:
So as mentioned in #9 heres the pull request for the review of the "prototype". Its verry basic atm.


src/partials/messenger.ts
```diff
@@ -79,6 +80,54 @@ class SendFileController extends DialogController {
...
+
+    protected hide(data: any): void {
+        this.$mdDialog.hide(data);
+        this.done();
+    }
+
+    private done(): void {
+        if (this.activeElement !== null) {
+            // reset focus
+            this.$window.location.reload();
```
dbrgn: Is there a way to reload the theme without reloading the entire application?

src/services/settings.ts
```diff
@@ -0,0 +1,114 @@
...
+            if (this.$window.localStorage === null) {
+                this.blocked = true;
+            }
+            this.storage = this.$window.localStorage;
+            // Load Initial Data from LocalStorage
+            this.currentTheme = this.retrieveUntrustedKeyValuePair('theme');
+            this.applyTheme();
+        } catch (e) {
+            this.$log.warn(this.logTag, 'LocalStorage blocked:', e);
+            this.blocked = true;
```
dbrgn: Threema Web checks for LocalStorage availability on the welcome screen, so you can assume it's available and can remove the error handling code.

src/partials/messenger.ts
FloThinksPi: I think so , there hase to be some way like the demo shows https://material.angularjs.org/1.1.0/demo/colors but i dont know how yet. I am looking more at applying theme colors consistent on all elements atm.

FloThinksPi:
So @dbrgn , i implemented the colors of the loginscreen to match the Theme with md-colors. We would have to add the md-colors tag to all affected elements. Unfortunally only Angular 2+ does support such stuff in CSS [See here](https://medium.com/covalent-ui/angular-material-2-theme-tutorial-2f7e6c344006#.5bq4pv2q2). But i still find the solution with md-colors quite nice, they can also get live updated like in the demo of AngularJS [here](https://material.angularjs.org/1.1.0/demo/colors)

dbrgn:
> Unfortunally only Angular 2+ does support such stuff in CSS

I didn't mean to rely on Angular for that :)

If you set a CSS class on the body tag:

```html
<body class="theme-dark">
```

...then you can add a scss file `src/sass/themes/_dark.scss` that looks like this:

```scss
body.theme-dark {
    h1.color {
        // ...
    }
    // ...
}
```

Other themes could be added as well. And the `theme-dark` class could be made dynamic by using a theme controller.

FloThinksPi:
closing this for now, will resubmit a pull request once i have it fully wokring 😉

PR #118, opened 2017-03-01
Enable TLS for TURN traffic

dbrgn:
Note that this is not security relevant. Content is always end-to-end encrypted. But it helps with DPI firewalls.

Refs #43

### DIFF
--- a/src/config.ts
+++ b/src/config.ts
@@ -19,7 +19,8 @@ export default {
     ICE_SERVERS: [{
         urls: [
             'turn:turn.threema.ch:443?transport=udp',
-            'turn:turn.threema.ch:443?transport=tcp'
+            'turn:turn.threema.ch:443?transport=tcp',
+            'turns:turn.threema.ch:443',
         ],
         username: 'threema-angular',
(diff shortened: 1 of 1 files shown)


rugk:
But is it in this case additionally TLS encrypted? And without it only the pure e2e encryption is used?

lgrahl:
No, it's actually encrypted three times - this is the stack (regarding cryptography):

SaltyRTC Secure Data Channel (NaCl, E2EE)
DTLS (whatever cypher suite has been chosen, E2EE)
TURN-TLS (whatever cypher suite has been chosen, transport encryption)

lgrahl:
When both peers use TURN-TLS relay candidates, it's even more insane. Then it will be TURN-TLS twice - one on both ends.

rugk:
> No, it's actually encrypted three times

Quite appropriate for **Three**ma I think… 😊🙉
Just kidding…

Indeed that's kinda insane. I would not say it has "no effect on security" - it is just a transport encryption, which is just unnecessary in this case. And I could think that this is really the maximum of encryption layers, otherwise there could be performance problems/battery drain, could not they?
I assume that's also the reason why the TURN-TLS is prioritized/tried last (I assume the array is ordered…), is not it?

Issue #76, opened 2017-02-21
Move from browser detection to feature detection


dbrgn:
## Current Behavior

When not using Chrom(e|ium) 45+, Firefox 47+ or Opera 32+, a warning shows up.

## Expected Behavior

When using any browser that supports our required features, no warning should show up.

## Possible Solution

Detect the following features:

- ES2015 (TODO: Which features exactly?)
- WebRTC PeerConnection
- WebRTC DataChannel

PR #126, opened 2017-03-02
Add troubleshooting page

dbrgn:
Screenshot:

diag

With JS disabled:

diag-nojs

Could be extended in the future with a network test.

Testing for ES2015 is both hard and should not be necessary. Browsers that provide WebRTC should already implement most of ES2015. And for others we have included a polyfill that provides features like promises.


dbrgn:
Refs #76

troubleshoot/index.html
```diff
@@ -0,0 +1,217 @@
...
+        function switchTo(type, newStatus) {
+            var unknown = document.querySelector('#status-' + type + ' .status-unknown');
+            if (unknown) {
+                unknown.classList.add('hidden');
+            }
+            document.querySelector('#status-' + type + ' .status-no').classList.add('hidden');
+            document.querySelector('#status-' + type + ' .status-yes').classList.add('hidden');
+            document.querySelector('#status-' + type + ' .status-' + newStatus).classList.remove('hidden');
+        }
+        function do_checks() {
```
sillych: missing camel case declaration :smile:

PR #127, opened 2017-03-02
Add TURN connectivity test to troubleshoot page

dbrgn:
Add a new section to the troubleshooting page:

turn-yes

This is the message that appears if connectivity fails:

turn-no

This should improve things for troubleshooting #43.


rugk:
From your screenshot: Why does not it also show tcp and TURN-HTTPS connection you added? Or does it do this only when the other connections fail?

dbrgn:
UDP specifies the "inner" protocol of the relayed packets, not the protocol between you and the TURN server.

It doesn't look like the browser provides us with information about whether TLS is enabled or not.

PR #129, opened 2017-03-03
Settings Menu

FloThinksPi:
So i removed theming so there is now an empty settings menu from pull request #110 for issue #8


dbrgn:
Thanks!

PR #130, opened 2017-03-03
Submit file using enter key

### LINKED ISSUE #122: Submit file using enter key
## Expected Behavior

When sending a file, the caption pop up appears. After typing a text, it should be possible to submit the file by pressing the enter key.

## Current Behavior

It doesn't work (at least in Firefox 45).

## Your Environment

* Threema Web version: 1.0.4
* Browser name and version: Firefox 45

IndianaDschones:
Fixes #122

[KeyboardEvent.charCode](https://developer.mozilla.org/en-US/docs/Web/API/KeyboardEvent/charCode) is deprechated. Used [KeyboardEvent.key](https://developer.mozilla.org/en-US/docs/Web/API/KeyboardEvent/key) instead.

### DIFF
--- a/src/partials/messenger.ts
+++ b/src/partials/messenger.ts
@@ -325,11 +325,11 @@ class ConversationController {
                         controllerAs: 'ctrl',
                         // tslint:disable:max-line-length
                         template: `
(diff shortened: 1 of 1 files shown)


dbrgn:
Thanks! Works as intended :)

Issue #108, opened 2017-02-28
Messages marked as seen in inactive session.


ranguwud:
## Expected Behavior
When the Window/Tab with Threema Web is not active/visible new incoming messages should not be marked as "seen" immediately.

## Current Behavior
If the conversation with contact X is selected in Threema Web and the Window/Tab is not active, all incoming messages from X will be marked as "seen" automatically. On the mobile device, any incoming message from X will create a sound/vibration and a notification is very briefly shown, but vanishes immediately after, thus leaving no trace of the unread message without opening the app. The status LED remains dark. This is particularly impractical e.g. when leaving the computer for a break and only locking its screen but not logging off Threema Web.

## Possible Solution
It should be possible to check, whether there is user interaction with Threema Web and mark incoming messages as "seen" only, if the user has focus on Threema Web. Possible approaches might be to check, whether the tab Threema Web resides in is active, or to watch for mouse moves/clicks or keystrokes sent to the tab. An option to fine-tune notifications on the mobile device, while working with Threema Web, as suggested in #21 might be helpful.

## Steps to Reproduce (for bugs)
1. In Threema Web, open conversation with contact X.
2. Leave Threema Web unattended, e.g. by switching tab, leaving the browser or locking the screen.
3. Have contact X send a message to you.
4. Notice that incoming messages are immediately marked as "seen" and that that the mobile device will vibrate/play a sound but that it won't show any notifications.

## Your Environment
* Threema Web version: 1.0.3
* Threema for Android version: 3.01.2000340
* Threema for Android source: Play Store
* Browser name and version: Firefox 51.0.1
* Operating system and version: Windows 7

ranguwud:
Issue #21 is related to this issue, but not a duplicate in my opinion. While I am missing notifications on the mobile device when on Threema Web, OP of #21 wants to turn off notifications on the mobile device. In general, of course, an option in the app to fine-tune the behavior would be most helpful.

dbrgn:
Related discussion: https://github.com/thenikso/angular-inview/issues/111

I'll give it a stab independently from the angular-inview implementation though.

ovalseven8:
Isn't this issue a duplicate of #17?

dbrgn:
> Isn't this issue a duplicate of #17?

No, it's a different issue (see #108).

dbrgn:
> Isn't this issue a duplicate of #17?

It's similar, but not the same technical cause.

dbrgn:
[...]

Issue #17, opened 2017-02-15
Threema Web does not necessarily show number of unread messages


ovalseven8:
Steps to reproduce:

1. Open https://web.threema.ch
2. Write a message to i.e. `ECHOECHO`
3. Switch to another (browser) tab *immediately*
4. You can see that the tab's title has changed from `Threema Web` to `Threema Web (1)`. However, because in this tab the chat with `ECHOECHO` is open, it will be marked as read. So, after about a second you read `Threema Web` again and you cannot see that you have unread messages. If the chat with `ECHOECHO` is not open, everything works as expected.

I am not sure if it's technically possible, but messages should only be marked as read when they are *really* read (in the sense of I look at the chat). Otherwise, I cannot see if I got any message.

rugk:
@dbrgn not "bug"?

luminatedhorizons:
Excellent description, thanks for that. Since in this scenario, incoming messages are marked as "read/seen" before even only the tab of Threema Web has been shown on the screen, I would maybe go as far to call this a "bug", no?

dbrgn:
To repeat my comment from #108:

Related discussion: https://github.com/thenikso/angular-inview/issues/111

PR #138, opened 2017-03-07
Hide desktop notifications for read conversations

### LINKED ISSUE #131: Also hide tab glow and possibly desktop notification on read
## Expected Behavior

When a message is read hide the desktop notification, if that's possible.
And when all messages are read, please also hide the tab glow, which is shown when a new message appears. Currently this glow is only hidden when you select the tab.

## Current Behavior

Once a message is received, then tab glow can only be hidden after clicking on the tab.

1. Get a message.
2. Read it on your phone.

## Your Environment

* Threema Web version: 1.0.5

dbrgn:
If the unread counter for a conversation decreases, hide related notifications.

Fixes #131.

### DIFF
--- a/src/services/notification.ts
+++ b/src/services/notification.ts
@@ -18,6 +18,8 @@
 import Receiver = threema.Receiver;
 export class NotificationService implements threema.NotificationService {
 
+    private logTag: string = '[NotificationService]';
+
     private $log: ng.ILogService;
     private $window: ng.IWindowService;
     private $state: ng.ui.IStateService;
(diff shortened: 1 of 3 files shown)


rugk:
However you should only release it when https://github.com/threema-ch/threema-web/issues/17/https://github.com/threema-ch/threema-web/issues/108 is fixed, because otherwise in the situation https://github.com/threema-ch/threema-web/issues/17 (and especially https://github.com/threema-ch/threema-web/issues/108) describes  also hides the desktop notification thoughtful it actually was not read.

Previously I could at least read the message in the desktop notifications in this case, but with this fix I assume that it will be hidden instantly.

PR #142, opened 2017-03-07
Show notification settings

### LINKED ISSUE #8: Show notification settings
Make it possible to enable/disable desktop notifications.

IndianaDschones:
Fixes #8 

I also added the ability to hide/show the preview text (like the Android App).

Todo: @dbrgn or someone at Threema should add a notification faq, because there are some pitfalls. I already added a link to https://threema.ch/de/faq/web_notifications in the settings dialog if the permission was rejected. Feel free to remove or change the url.

For example if the user permanently blocked notifications for Threema Web he must manually re-grant permission via browser settings or by clicking on the icon at the url-bar and must reload the page.
Some browsers, for example firefox, offer to mute notifications until restarting the browser. Problem: this can´t be detected. So the user may think "I checked the notification checkbox and granted permission" but no notification shows up.

### DIFF
--- a/public/i18n/de.json
+++ b/public/i18n/de.json
@@ -164,7 +164,7 @@
         "TEXT_TOO_LONG": "Diese Nachricht ist zu lang und kann nicht gesendet werden (Maximale Länge {max} Zeichen).",
         "NOTIFICATION_PERMISSION_DENIED" : "Berechtigung nicht erteilt. Sie müssen die Berechtigungen für Threema Web manuell erteilen.",
         "NOTIFICATION_PERMISSION_DENIED_LEARN_MORE" : "Mehr erfahren.",
-        "API_NOT_AVAILABLE" : "Ihr Browser unterstütz keine Desktopbenachrichtigungen."
+        "API_NOT_AVAILABLE" : "Ihr Browser unterstützt keine Desktopbenachrichtigungen."
     },
     "mimeTypes": {
         "android_apk": "Android-Paket",
--- a/public/i18n/de.json
+++ b/public/i18n/de.json
@@ -164,7 +164,7 @@
         "TEXT_TOO_LONG": "Diese Nachricht ist zu lang und kann nicht gesendet werden (Maximale Länge {max} Zeichen).",
         "NOTIFICATION_PERMISSION_DENIED" : "Berechtigung nicht erteilt. Sie müssen die Berechtigungen für Threema Web manuell erteilen.",
         "NOTIFICATION_PERMISSION_DENIED_LEARN_MORE" : "Mehr erfahren.",
-        "API_NOT_AVAILABLE" : "Ihr Browser unterstützt keine Desktopbenachrichtigungen."
+        "NOTIFICATION_API_NOT_AVAILABLE" : "Ihr Browser unterstützt keine Desktopbenachrichtigungen."
     },
     "mimeTypes": {
         "android_apk": "Android-Paket",
@@ -198,7 +198,7 @@
         "notifications": {
             "NOTIFICATIONS" : "Benachrichtigungen",
             "SHOW_NOTIFICATIONS" : "Desktopbenachrichtigungen anzeigen",
-            "SHOW_PREVIEW" : "Vorschau anzeigen"
+            "SHOW_PREVIEW" : "Nachrichteninhalt in Benachrichtigungen anzeigen"
         }
     }
 }
--- a/public/i18n/de.json
+++ b/public/i18n/de.json
@@ -191,6 +191,10 @@
         "LICENSE_LINK_AFTER": "gefunden werden"
     },
     "settings": {
-        "SETTINGS":"Einstellungen"
+        "SETTINGS":"Einstellungen",
+        "notifications": {
+            "NOTIFICATIONS" : "Benachrichtigungen",
+            "SHOW_NOTIFICATIONS" : "Desktopbenachrichtigungen anzeigen"
+        }
     }
 }
--- a/public/i18n/de.json
+++ b/public/i18n/de.json
@@ -161,7 +161,10 @@
         "FILE_MESSAGES_NOT_SUPPORTED": "«{receiverName}» kann noch keine Dateien erhalten.",
         "ERROR_OCCURRED": "Es ist ein Fehler aufgetreten.",
         "FILE_TOO_LARGE": "Aktuell können keine Dateien grösser als 15 MiB über Threema Web versendet werden",
-        "TEXT_TOO_LONG": "Diese Nachricht ist zu lang und kann nicht gesendet werden (Maximale Länge {max} Zeichen)."
+        "TEXT_TOO_LONG": "Diese Nachricht ist zu lang und kann nicht gesendet werden (Maximale Länge {max} Zeichen).",
+        "NOTIFICATION_PERMISSION_DENIED" : "Berechtigung nicht erteilt. Sie müssen die Berechtigungen für Threema Web manuell erteilen.",
+        "NOTIFICATION_PERMISSION_DENIED_LEARN_MORE" : "Mehr erfahren.",
+        "API_NOT_AVAILABLE" : "Ihr Browser unterstütz keine Desktopbenachrichtigungen."
     },
     "mimeTypes": {
         "android_apk": "Android-Paket",
@@ -194,7 +197,8 @@
         "SETTINGS":"Einstellungen",
         "notifications": {
             "NOTIFICATIONS" : "Benachrichtigungen",
-            "SHOW_NOTIFICATIONS" : "Desktopbenachrichtigungen anzeigen"
+            "SHOW_NOTIFICATIONS" : "Desktopbenachrichtigungen anzeigen",
+            "SHOW_PREVIEW" : "Vorschau anzeigen"
         }
     }
 }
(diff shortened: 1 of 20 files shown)


src/services/notification.ts
```diff
@@ -16,55 +16,170 @@
...
 
+    private static SETTINGS_NOTIFICATIONS = 'notifications';
+    private static SETTINGS_NOTIFICATION_PREVIEW = 'notificationPreview';
+
     private $log: ng.ILogService;
     private $window: ng.IWindowService;
     private $state: ng.ui.IStateService;
 
+    private settingsService: SettingsService;
+    private logTag = '[Notification Service]';
```
dbrgn: Can you leave away the whitespace in here, to match the class name?

src/services/webclient.ts
```diff
@@ -407,8 +407,9 @@ export class WebClientService implements threema.WebClientService {
             this.$log.debug('Handover done');
 
             // Ask user for notification permission
-            this.$log.debug('Check notification permission...');
-            this.notificationService.requestNotificationPermission();
+            this.$log.debug('Check notification API and fetch initial settings...');
+            this.notificationService.checkNotificationAPI();
+            this.notificationService.fetchSettings();
```
dbrgn: Since these methods should always be called together, could you make them private and add a new public function `notification.Service.init()` that calls those other two methods? Might prevent bugs in the future.

src/services/notification.ts
```diff
@@ -16,55 +16,170 @@
...
+        this.settingsService = settingsService;
     }
 
     /**
      * Ask user for desktop notification permissions.
      *
      * Updates internal `maxNotify` flag. If the user accepts, the is set to to
      * true. If the user declines, the flag is set to false. If the
-     * Notifications API is not available, the flag is set to null.
+     * user makes no choice, the flag is set to null
```
dbrgn: Can you update the comment with the correct flag variables? `mayNotify` (there was a typo in there) doesn't exist anymore.

src/services/notification.ts
```diff
@@ -16,55 +16,170 @@
...
+     * Denied: Don´t ask again, user must manually re-grant permission
+     * via browser settings
+     * Granted: We have the (browser-)permission to send notifications
+     * Default: Initially state on first visit
+     */
+    public checkNotificationAPI(): void {
+        this.notificationAPIAvailable = ('Notification' in this.$window);
+        this.$log.debug(this.logTag, 'Notification API available:', this.notificationAPIAvailable);
+        const Notification = this.$window.Notification;
+        switch (Notification.permission) {
```
dbrgn: You'll get a TypeError here if the notification API is not available.

In theory we don't support browsers without a desktop notification system, but there may also be strange configurations around, or people with funny plugins blocking stuff.

public/i18n/de.json
```diff
@@ -191,6 +194,11 @@
         "LICENSE_LINK_AFTER": "gefunden werden"
     },
     "settings": {
-        "SETTINGS":"Einstellungen"
+        "SETTINGS":"Einstellungen",
+        "notifications": {
+            "NOTIFICATIONS" : "Benachrichtigungen",
+            "SHOW_NOTIFICATIONS" : "Desktopbenachrichtigungen anzeigen",
+            "SHOW_PREVIEW" : "Vorschau anzeigen"
```
dbrgn: Maybe it would be clearer to the user what this does if it said "Nachrichteninhalt in Benachrichtigung anzeigen"? And in English, "Show message contents in notifications".

public/i18n/de.json
```diff
@@ -161,7 +161,10 @@
         "FILE_MESSAGES_NOT_SUPPORTED": "«{receiverName}» kann noch keine Dateien erhalten.",
         "ERROR_OCCURRED": "Es ist ein Fehler aufgetreten.",
         "FILE_TOO_LARGE": "Aktuell können keine Dateien grösser als 15 MiB über Threema Web versendet werden",
-        "TEXT_TOO_LONG": "Diese Nachricht ist zu lang und kann nicht gesendet werden (Maximale Länge {max} Zeichen)."
+        "TEXT_TOO_LONG": "Diese Nachricht ist zu lang und kann nicht gesendet werden (Maximale Länge {max} Zeichen).",
+        "NOTIFICATION_PERMISSION_DENIED" : "Berechtigung nicht erteilt. Sie müssen die Berechtigungen für Threema Web manuell erteilen.",
+        "NOTIFICATION_PERMISSION_DENIED_LEARN_MORE" : "Mehr erfahren.",
+        "API_NOT_AVAILABLE" : "Ihr Browser unterstützt keine Desktopbenachrichtigungen."
```
dbrgn: This should be `NOTIFICATION_API_NOT_AVAILABLE` to match the string actually used. Same for the english file.

src/partials/dialog.settings.html
```diff
@@ -11,9 +11,47 @@ <h2 translate>settings.SETTINGS</h2>
...
+                                    ng-change="ctrl.setWantsPreview(ctrl.notificationPreview)"
+                                    aria-label="Show message preview">
+                            </md-checkbox>
+                            <span translate>settings.notifications.SHOW_PREVIEW</span>
+                        </md-list-item>
+                    </md-list>
 
+                    <div ng-if="!ctrl.notificationApiAvailable">
+                        <span translate>error.NOTIFICATION_API_NOT_AVAILABLE</span>
+                    </div>
```
dbrgn: I would actually hide the entire settings options if the API is not available. There's no point in showing options to the user that have no effect. Especially since the first option is visible and checked, but greyed out if the API is not available.

src/partials/dialog.settings.html
```diff
@@ -11,9 +11,47 @@ <h2 translate>settings.SETTINGS</h2>
...
+                                    ng-disabled="!ctrl.notificationApiAvailable || ctrl.notificationPermission === false"
+                                    ng-model="ctrl.desktopNotifications"
+                                    ng-change="ctrl.setWantsNotifications(ctrl.desktopNotifications)"
+                                    aria-label="Show desktop notifications">
+                            </md-checkbox>
+                            <p translate>settings.notifications.SHOW_NOTIFICATIONS</p>
+                        </md-list-item>
+                        <md-list-item>
+                            <md-checkbox
+                                    ng-disabled="!ctrl.desktopNotifications"
```
dbrgn: There's a little usability issue here: If the notifications are disabled (first option) and the preview is disabled too (second option), it's barely visible that the option is greyed out and cannot be checked.

Maybe we can hide the entire option if notifications are disabled?

public/i18n/de.json
IndianaDschones: I used the strings provided by the Android App. I think it would be better to have consistent strings for the same setting(s) across the platforms.

But if you want i can change them, your decision ;)

src/partials/dialog.settings.html
IndianaDschones: Yeah, that might be better. I´ll change that :)

src/partials/dialog.settings.html
IndianaDschones: It finally works as intended. At first i had the string inside `md-checkbox` but it didn´t show, so i put it outside. But according to the docs it shoult work this way. The problem was that angular thougt it would be okay to set the visibilit of the label to `display: none`... 😑
I added a custom CSS rule to show them

Will create a PR in a few

public/i18n/de.json
dbrgn: Yes, we don't necessarily need to match the Android app strings.

src/partials/dialog.settings.html
IndianaDschones: Shuld i also hide the options if the permission wasn´t granted? The `NOTIFICATION_PERMISSION_DENIED`-Hint shows anyway and should be enough.

What do you think @dbrgn ?

dbrgn: That's nice, and a fix for the general case is welcome. But in this case I actually think it's better to hide the option completely with ng-if, don't you think?

@sillych what's your opinion?

dbrgn: > Should i also hide the options if the permission wasn´t granted?

Yes, I agree about the hint. It's sufficient.

If you want, you could prepend the "error" material icon and make it red with CSS (color: $status-error).

IndianaDschones: > If you want, you could prepend the "error" material icon to the messsage and make it red 

I copied your css `.status span`, `.status-yes/no/unknown i` from the troubleshooting page into the `_main.scss`. May come in handy somewhere else ;)

dbrgn:
There's a conflict with `notification.ts` on master. Could you rebase with master so we can merge?

By the way, there's a usability improvement that we should probably do. Right now the user can set the notification permission in the settings, then ignore the permission (or click on "not now") and then click on "ok" in the settings dialog. No warning appears, but when opening the dialog again the option disabled.

Instead, the user should probably be shown a dialog letting him/her know that the permission hasn't been granted so far.

But that change can happen separately from this PR.

IndianaDschones:
I´m not sure if the rebasing worked as intended... 😕

IndianaDschones:
Now it looks better - I think. The conflict with `notification.ts` is gone.

If i did something wrong let me know ;)

dbrgn:
It looks a bit strange - the PR somehow contains duplicate commits now.

It's no problem though, I still have a checkout of the previous - and tested - state on a computer. I'll merge this PR on Tuesday using the "good" state, when I have access to that computer again. (There won't be a release before that day anyways.)

IndianaDschones:
Closed in favor of PR #155

PR #145, opened 2017-03-08
Show group members in header

### LINKED ISSUE #6: Show group members in header
Show group members in header, below group name (similar to app).

IndianaDschones:
Fixes #6 

- Also added the "truncate-behaviour" to the receiver name.
- Added missing translation in about dialog

### DIFF
--- a/src/filters.ts
+++ b/src/filters.ts
@@ -244,4 +244,18 @@ angular.module('3ema.filters', [])
     };
 }])
 
+/**
+ * Convert ID-Array to (Display-)Name-Array
+ */
+    .filter('idToName', ['WebClientService', function (webClientService: threema.WebClientService) {
+    return(ids: string[]) => {
+        let names: string[] = [];
+        for (let id of ids){
+            this.contactReceiver = webClientService.contacts.get(id);
+            names.push(this.contactReceiver.displayName);
+        }
+        return names;
+    };
+}])
+
 ;
--- a/src/filters.ts
+++ b/src/filters.ts
@@ -245,7 +245,7 @@ angular.module('3ema.filters', [])
 }])
 
 /**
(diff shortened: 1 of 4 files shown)


src/filters.ts
```diff
@@ -244,4 +244,18 @@ angular.module('3ema.filters', [])
     };
 }])
 
+/**
+ * Convert ID-Array to (Display-)Name-String, separated by ','
+ */
+    .filter('idToName', ['WebClientService', function (webClientService: threema.WebClientService) {
```
dbrgn: The `.filter` call should not be indented, to match the rest of the file.

Also, the function should probably be named `idsToNames`, since it accepts multiple ids.

src/sass/sections/_conversation.scss
```diff
@@ -29,6 +29,14 @@
             padding: 0;
             font-size: 120%;
         }
+
+        .conversation-header-details-detail, .conversation-header-details-name {
+            white-space: nowrap;
+            overflow: hidden;
+            text-overflow: ellipsis;
+            max-width: 100%;
+            display: inherit;
```
dbrgn: Can you add `line-height: 1.3` here? Otherwise text can be cut off.

src/partials/messenger.conversation.html
```diff
@@ -20,6 +20,9 @@
             <eee-verification-level ng-if="ctrl.type == 'contact'"
                                     contact="ctrl.receiver"></eee-verification-level>
             </div>
+            <div class="conversation-header-details-detail" ng-if="ctrl.type == 'group'">
```
sillych: please add the whole list to the `title` attribute

    title="{{ctrl.receiver.members | idToName}}"

dbrgn:
Thanks!

Issue #121, opened 2017-03-02
Support Firefox ESR


dbrgn:
## Expected Behavior

It would be good to officially support the extended support release (ESR) of Firefox, which is currently at version 45.

## Current Behavior

Firefox 45 is not supported so far because we didn't test it. However, I see no reasons why it shouldn't be possible. It seems to work fine.

## Possible Solution

After testing, simply set `BROWSER_MIN_VERSIONS.FF` from 47 to 45 in `src/app.ts`.

dbrgn:
Pasting does not work in Firefox 45.

rugk:
Should not this be fixed when you switch to feature detection?

dbrgn:
As you see here:

> Pasting does not work in Firefox 45.

...there are some things that can't easily be detected. We need more testing before we can remove the warnings. Decreasing `BROWSER_MIN_VERSIONS.FF` would be an intermediate step.

XS666:
Please fix this... I am using Firefox ESR on all my computers.

dbrgn:
Blocked by #139.

@XS666 in the meantime, you can just ignore the warning. Most things should work.

dbrgn:
Another blocker: Firefox ESR never notices if the WebRTC connection is lost. Connection is still reported as "up" in `about:webrtc`. Attempting to send data through the DC doesn't make a difference.

Steps to reproduce: Create a connection, kill the Threema app.

Not sure if this is something we can handle. If the bug persists, I don't see how we can support Firefox ESR.

dbrgn:
Since the WebRTC connection bug persists, we will not be able to properly support Firefox 45. But the good news is that Firefox ESR 52 was released yesterday! https://www.mozilla.org/en-US/firefox/organizations/faq/

This means that we automatically support the current version of Firefox ESR, as of now. I would expect distributions like Debian to update their packages soon.

PR #147, opened 2017-03-08
simplify avatar html code

### LINKED ISSUE #143: Avatar flickering
Adding the [ng-cloak directive to avatars](https://github.com/threema-ch/threema-web/commit/afbc49ba3280de2e9242aae9eaa3da082017a060) causes them to flicker when being reloaded.

sillych:
- remove default avatar image tag
- remove ng-cloak
- request avatar only if the avatar is more than 0.5s inview (do not load all avatars on scrolling down the contact/conversation list)
- fix #143

### DIFF
--- a/src/directives/avatar.ts
+++ b/src/directives/avatar.ts
@@ -51,7 +51,7 @@ export default [
                 this.getAvatar = () => {
                     if (this.avatarExists()) {
                         return this.receiver.avatar[this.resolution];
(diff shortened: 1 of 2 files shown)


dbrgn:
Cool! It fixes flickering, but unfortunately the avatar loading bug in Firefox 45 still occurs... :(

On the other hand, we can now probably ignore that problem, since the new Firefox ESR version is 52 (see #121). What do you think @sillych?

sillych:
@dbrgn deal!

PR #150, opened 2017-03-09
Check for desktop notifications in troubleshooting tool

dbrgn:


dbrgn:
Maybe we should also check whether the user has granted them

PR #165, opened 2017-03-13
Markup behaves different than app

### LINKED ISSUE #156: Markup behaves different than app
## Expected Behavior

Markup should behave identical to the Android app.

## Current Behavior

The app does not apply formatting to multi line strings, web does.

Example:

```
*foo bar
baz cyberkrimi*
```

This is bold in web but not in the app.

## Possible Solution

We should probably do it like the app, since a user on a mobile will probably not use manual line breaks when creating formatted text.

The regex actually matches for `[^\r\n]`, but newslines are apparently still matched. Maybe this is due to the conversion of newlines to HTML, or simply a JavaScript idiosyncracy.

Adding tests for this would be good, to prevent regressions.

IndianaDschones:
Fixes #156 
Markup should now behave like the (Android) app. Also added some tests.

### DIFF
--- a/src/directives/message_quote.ts
+++ b/src/directives/message_quote.ts
@@ -36,7 +36,7 @@ export default [
             template: `
                 <div class="message-quote-content" ng-style="{'border-color': ctrl.contact.color}">
                     <span class="message-name" ng-style="{'color': ctrl.contact.color}">{{ ctrl.contact.displayName }}</span>
-                    <span class="message-quote" ng-bind-html="ctrl.text | escapeHtml | nlToBr | emojify | linkify | markify"></span>
+                    <span class="message-quote" ng-bind-html="ctrl.text | escapeHtml | markify | nlToBr | emojify | linkify"></span>
                 </div>
             `,
         };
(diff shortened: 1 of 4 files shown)


dbrgn:
Hm, that seems like an invitation for future bugs to happen.

Wouldn't it be the better solution to simply change the order of filters [in the directive](https://github.com/threema-ch/threema-web/blob/master/src/directives/message_text.ts#L51)?

I think this should work:

    ctrl.text | escapeHtml | markify | emojify | linkify | nlToBr: ctrl.multiLine

IndianaDschones:
Good point 👍. Haven´t thought about the order ^^
Updating PR in a few

src/directives/message_quote.ts
```diff
@@ -36,7 +36,7 @@ export default [
             template: `
                 <div class="message-quote-content" ng-style="{'border-color': ctrl.contact.color}">
                     <span class="message-name" ng-style="{'color': ctrl.contact.color}">{{ ctrl.contact.displayName }}</span>
-                    <span class="message-quote" ng-bind-html="ctrl.text | escapeHtml | nlToBr | emojify | linkify | markify"></span>
+                    <span class="message-quote" ng-bind-html="ctrl.text | escapeHtml | markify | nlToBr | emojify | linkify"></span>
```
dbrgn: Can you move the `nlToBr` filter to the end, so that the order is the same as in `message_text.ts`?

IndianaDschones:
Here you go :)

PR #167, opened 2017-03-13
Fix / improve soft reconnect behavior

dbrgn:
This adds two main fixes / improvements:

- Previously, the initialization steps were not properly reset after a full intiialization. This meant that a soft reconnect would always fail, because the WebClientService would ignore already registered intiialization steps. This PR adds proper reset calls.
- Previously, with a timeout of 20 seconds, it could happen that a connection process would time out even though it was still in process (e.g. if the push takes a while to arrive). This PR resets the timeout every time an important connection buildup step is reached.

For the connection buildup state tracking, a new event emitter library was added: [ts-events](https://github.com/rogierschouten/ts-events). It allows adding self-contained event emitters as class attributes. Maybe we can convert some watches to explicit event emitters in the future. It provides a clear API and should improve performance compared to AngularJS watches.

Refs #106 (there are some other things that should be improved before it can be considered fixed).

### DIFF
--- a/LICENSE-3RD-PARTY.txt
+++ b/LICENSE-3RD-PARTY.txt
@@ -707,7 +707,7 @@ License for saltyrtc-client
 
 The MIT License (MIT)
 
-Copyright (c) 2016 Threema GmbH
+Copyright (c) 2016-2017 Threema GmbH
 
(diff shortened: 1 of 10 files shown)


dbrgn:
✨

PR #169, opened 2017-03-13
Translated meta description

### LINKED ISSUE #168: Translate meta description
The `meta description` added in #124 should also be translated.

I self assigne me.

IndianaDschones:
Fixes #168


dbrgn:
There should probably be a fallback description in English, in case the crawler doesn't enable JavaScript.

As far as I know, Angular will simply overwrite the existing value if it exists already.

PR #178, opened 2017-03-16
Audio notification for new messages

### LINKED ISSUE #16: Audio notification for new messages
At the moment, I only get notified about new messages when the desktop notification option is enabled.

However, it does not play any sound on my computer (Firefox 51.0.1; Ubuntu 16.04).
It would be great if Threema Web played a sound when you get a new message (and independent from desktop notification).

Of course it should be possible to enable/disable the sound notification.

IndianaDschones:
Fixes #16 

A notification sound can be played independent from the notification permission/setting.

**Todo (someone @threema-ch or @dbrgn)**

- [x] Upload notification sound (MP3)


src/services/notification.ts
```diff
@@ -213,6 +233,16 @@ export class NotificationService {
     }
 
     /**
+     * Sets if the user wants sound when a new message arrives
+     * @param wantsSound
+     */
+    public setWantsSound(wantsSound: boolean): void {
+        this.$log.debug(this.logTag, 'Requesting sound preference change to', wantsSound);
+        this.notificationSound = wantsSound;
+        this.storeSetting(NotificationService.SETTINGS_NOTIFICATION_SOUND, wantsSound.toString());
```
dbrgn: I'd probably do an explicit string conversion here instead of relying on `toString`, just to be on the safe side.

sillych:
Good work, Thanks! I will push some changes (like missing sound file) into your branch

src/services/notification.ts
IndianaDschones: Okay, using String(value) now :)

dbrgn:
@IndianaDschones sorry, I meant something like `wantsPreview ? 'true' : 'false'`, since you also compare with those values when reading them back.

IndianaDschones:
Okay I´ll change it during the day :)

sillych:
Thank you!

PR #185, opened 2017-03-23
Location messages redesigned

### LINKED ISSUE #181: Location link can not be clicked
## Expected Behavior

Clicking on a location-link should work
## Current Behavior

Link is not reachable/can not be clicked.
## Solution

Hide the `in-view-indicator` on locations

I´ll solve this.

IndianaDschones:
Fixes #181 
- Location links can now be clicked
- Redesigned location messages (look now like media messages)
The location poi text (here "Reichstagsgebäude") is needed for the message preview (recent chats)
locationnew
- Added missing translation "Ballot messages not supported"


dbrgn:
Thanks, looks great!

Two issues:

- Since hovering over it also shows the hand mousepointer, can you make the icon clickable too? The best way to do this would probably be to move the `ctrl.location | mapLink` expression to a controller method and to use the ng-click directive on the icon. You can call the filter from the controller using `$filter('mapLink')(location)`.
- For some reason, neither the text below the icon/link nor the timestamp are selectable with the mouse. The mouse cursor does not change on hover either. Maybe an overlapping invisible div?

IndianaDschones:
Okay. The "clicking"-Issue also appears on the other media messages. I try to harmonize it (eg clicking on the icon and the text works or only the text) 😄

dbrgn:
> The "clicking"-Issue also appears on the other media messages

Thanks, I wasn't aware of that :)

PR #193, opened 2017-04-02
Enhanced desktop notification preview

### LINKED ISSUE #186: Enhance desktop notification preview
## Expected Behavior

When sending non-text messages (eg locations) the desktop notification preview should show
`Location: Reichstagsgebäude`
The media type should be translated.

## Current Behavior

`[location]`

## Solution
Implement it this way.

I´d like to fix this :)

IndianaDschones:
Fixes #186 

- Desktop notifications now include captions, filenames and translated filetype:
vorschau

- Gifs now show the 'image' icon in the recent list (like the Android app)

@sillych Can you compare my implementation with the Android implementation? Maybe I forgot some filetype. Asking is easier than trial and error ;)

### DIFF
--- a/src/services/webclient.ts
+++ b/src/services/webclient.ts
@@ -2011,11 +2011,10 @@ export class WebClientService {
                 let messageType = message.type;
                 let caption = message.caption;
                 let captionString = '';
-                if (typeof caption !== "undefined"){
+                if (typeof caption !== 'undefined') {
                     captionString = captionString + ': ' + caption;
                 }
                 let messageTypeString = this.$translate.instant('messageTypes.' + messageType);
-                console.info(messageTypeString + ": " + messageType);
                 switch (messageType) {
                     case 'text':
                         body = message.body;
--- a/src/services/webclient.ts
+++ b/src/services/webclient.ts
@@ -29,6 +29,7 @@ import {QrCodeService} from './qrcode';
 import {ReceiverService} from './receiver';
 import {StateService} from './state';
 import {TitleService} from './title';
+import {log} from "util";
 
 class WebClientDefault {
     private avatar: threema.AvatarRegistry = {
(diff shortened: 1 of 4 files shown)


src/services/webclient.ts
```diff
@@ -2008,12 +2008,39 @@ export class WebClientService {
             .then((titlePrefix) =>  {
                 const title = `${titlePrefix} ${senderName}`;
                 let body = '';
-                switch (message.type) {
+                let messageType = message.type;
+                let caption = message.caption;
+                let captionString = '';
+                if (typeof caption !== 'undefined') {
```
dbrgn: You can use `caption !== undefined` here. The `undefined` object is immutable since ES5, so it should be safe to compare to it directly.

src/services/webclient.ts
```diff
@@ -2008,12 +2008,39 @@ export class WebClientService {
...
+                        body = messageTypeString + ': ' + message.location.poi;
+                        break;
+                    case 'file':
+                        if (message.file.type === 'image/gif') {
+                            body = this.$translate.instant('messageTypes.' + 'gif') + captionString;
+                            break;
+                        }
+                        // Display caption, if available otherwise use filename
+                        if (captionString.length > 0) {
+                            body = messageTypeString + captionString;
```
dbrgn: Isn't a `': '` separator missing here? (Same on line 2027...)

src/services/webclient.ts
```diff
@@ -2008,12 +2008,39 @@ export class WebClientService {
...
+                            break;
+                        }
+                        body = messageTypeString + ': ' + message.file.name;
```
dbrgn: Since this is an if-else situation, it would probably be clearer if you put this part of the logic into an `else` block. Then you can also remove the `break` online 233.

src/services/webclient.ts
```diff
@@ -2008,12 +2008,39 @@ export class WebClientService {
...
+                        break;
+                    case 'ballot':
+                        // TODO Show ballot title if ballot messages are implemented in the web version
+                        body = messageTypeString;
+                        break;
                     default:
-                        body = `[${message.type}]`;
+                        // Image, video and audio
+                        body = messageTypeString + captionString;
```
dbrgn: Again, I think the separator (`': '`) is missing?

src/services/webclient.ts
IndianaDschones: It´s not missing it´s outsourced :)

On line 2013 `captionString` is initialized as an empty string. In the next few lines if the media has a caption it is added along with `': '`. The `captionString` will look like `: Your caption here` and the body will be something like `Image: Your caption here`.
If there is no caption it will be `Image`.

src/services/webclient.ts
IndianaDschones: Okay, I will change that and the `undefined` issue

src/services/webclient.ts
dbrgn: Ah, I see. I missed that you overwrote `captionString` oin line 2015 :)

dbrgn:
Thanks!

PR #203, opened 2017-04-26
fixed little typo

Pythonix:
I changed EmoijPickerContainer  to EmojiPickerContainer, which is the correct spelling.


dbrgn:
Hi @Pythonix. Thanks for your contribution!

I think we can skip the CLA stuff this time, since this is just a typo fix and not new code :) The error message shown aboved is caused by using a different e-mail address for your git commit than for your Github account.

PR #204, opened 2017-04-27
[UX] enforce 'cursor: text' on compose box

econic:
In Chrome when the compose box is focused, only its padding has a text cursor, the containing ::before element has a default cursor.
This does not apply in FF.
Alternatively you could define the text cursor for the :before element, but considering this would have to be set on all future child elements, setting it on the parent is prabably be more future proof.


dbrgn:
That's a good point, thanks for the fix! 👍

PR #210, opened 2017-05-01
Only change read status if in the same tab

Pythonix:
I changed the code, so that the read status of the message is only changed if the tab with the chat opened is the active browser window.

### DIFF
--- a/src/directives/compose_area.ts
+++ b/src/directives/compose_area.ts
@@ -82,7 +82,7 @@ export default [
                 /**
                  * Stop propagation of click events and hold htmlElement of the emojipicker
                  */
-                let EmoijPickerContainer = (function() {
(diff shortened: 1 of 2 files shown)


dbrgn:
Hi

Thanks for your contribution! I assume that this intends to fix #17.

We already have a pull request open against one of the libraries we use to detect whether a message is in the viewport or not: https://github.com/thenikso/angular-inview/pull/122 It has some advantages over your solutions, for example that it doesn't require setting timeouts and that the `visibilityState` attribute is also considered.

Since that pull request hasn't been merged for almost two months now, I'll vendorize the angular-inview library with the fix until something happens upstream. I'll close your PR, but will create another PR soon with a fix for #17.

PR #215, opened 2017-05-04
Support all threema url actions

### LINKED ISSUE #190: Sharer-URL
Are there any plans to implement a simple Sharer-URL for the web-client like whatsapp and other messengers already have?

`https://web.whatsapp.com/send?text=TEXTorURL` (web app)

`whatsapp://send?text=TEXTorURL` (mobile)

This snippet only works on mobile devices:

`threema://compose?text=TEXTorURL`

I'm looking forward to it. :-)

Best regards from Berlin
Sebastian

dbrgn:
- Add `UriService` with `parseQueryParams` function
- Refactor `ThreemaAction` directive
- Add support for `compose` directive

I tested various combinations of predefined compose text, drafts, clicking on back button and clicking on other conversations. I hope I didn't miss some edge case. If a draft is defined, it's overwriten, but when pushing the back button in the header without typing anything, the draft should be restored when going back to the conversation.

Fixes #190.


rugk:
So how did you finally do it? (Just give the ones an overview, who do not want to dive into the code.)
Using WebWorkers or some other messaging method or how exactly?
Or do you support `web-threema://` URIs or so?

IndianaDschones:
@rugk The url actions are only supported/handled **inside** Threema Web for sent and received messges.

I also tried to click threema://add-Links on different websites and nothing happened. Then i read the issue again, especially [this comment from dbrgn](https://github.com/threema-ch/threema-web/issues/190#issuecomment-299186607) where he said

> I'll add support for the compose action from within Threema Web though.

src/directives/threema_action.ts
```diff
@@ -14,38 +14,80 @@
...
+                    id: params.id,
+                    initParams: {
+                        text: text,
+                    },
+                });
+            };
+        };
+
+        const getThreemaActionHandler = (name: string) => {
+            switch (name) {
```
IndianaDschones: Query parameters should be case-insensitve 😉 
This could be solved using `name.toLowerCase()`

src/directives/threema_action.ts
```diff
@@ -14,38 +14,80 @@
...
-                                            type: 'contact',
-                                            initParams: {
-                                                identity: threemaId,
-                                            },
-                                        });
-                                    });
+                            && node.tagName.toLowerCase() === 'a') {
+
+                            const link = (node as HTMLElement).innerText;
+                            if (link !== undefined && link.startsWith('threema://')) {
```
IndianaDschones: Same as my first comment. Protocol should also be case-insensitive

rugk:
Great, that's at least somehow, although it of course does not fulfil https://github.com/threema-ch/threema-web/issues/190 completely.

PR #226, opened 2017-05-17
Version checking

### LINKED ISSUE #183: WebUI does not check for new version
## Expected Behavior

The UI should warn the user from using an outdated version of the WUI.

## Current Behavior

no_update_check

## Possible Solution

- Cache more, so it still works without the update
- Ask the user to refresh the page and load the new version

1. Use WUI, eg. 1.0.5
2. Wait a bit until a new version gets deployed, eg. 1.1.0
3. See the funny picture from current behavior

## Your Environment

* Threema Web version: 1.0.5 -> 1.1.0
* Threema for Android version: the latest to this day
* Browser name and version: Chromium 57.0.2987.110 (64-bit)
* Operating system and version: Arch Linux 2017-03-18

dbrgn:
On connect, initialize the version by requesting the `version.txt` file.

On every reconnect and on every conversation load, check for the version update.

If the version has changed, show this dialog:

new-version

Once the user presses "OK", the page is reloaded.

Fixes #183.


ovalseven8:
Why the check and request on every conversation change?

dbrgn:
Because loading the conversation loads a template via HTTP. I'm not sure how much caching there is being done by Angular, but by assuming that the template is re-loaded we're on the safe side. That template might belong to the new version and might therefore be broken when included from an old version.

PR #240, opened 2017-05-31
Right-to-left support

### LINKED ISSUE #239: Cannot send text in Right-to-Left mode
## Current Behavior

Text in right-to-left mode cannot be sent.

1. Open a conversation Threema Web in Chrome
2. Right-click into the compose area, choose right-to-left mode
3. Enter some text
4. Press enter, a toast "An error occured" will appear

dbrgn:
This adds support for sending right-to-left text.

Open issue: Inserting emojis replaces the nested div.

Fixes #239

### DIFF
--- a/src/directives/compose_area.ts
+++ b/src/directives/compose_area.ts
@@ -120,28 +120,36 @@ export default [
                 function submitText(): Promise<any> {
                     let text = '';
 
-                    // Extract text
-                    // tslint:disable-next-line: prefer-for-of (see #98)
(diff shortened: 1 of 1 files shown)


dbrgn:
Emoji are fixed.

Open issue: The placeholder text does not appear in Chrome if right-to-left or left-to-right mode are explicitly chosen (due to the nested div). But I think that has priority. We could open a separate "easy" issue, maybe someone else wants to give it a stab.

PR #254, opened 2017-06-21
Receiver editing: Only send avatar if changed

dbrgn:
This is a follow-up pull request for #244.

It needs a protocol version update, since the current app version requires the avatar field. Therefore, don't merge yet!


dbrgn:
Rebased and merged into `ios` branch (protocol v2).

Issue #39, opened 2017-02-15
Very high cpu load while using webclient


rohlandm:
While using the web client, I have a very high CPU load, using both Google Chrome and Vivaldi (Chromium based). I have attached a screenshot of the chrome task manager:

![threema_cpu](https://cloud.githubusercontent.com/assets/2794686/23000133/56d702c8-f3df-11e6-894f-3b741db1718b.png)

This issue hinders me from using the web client in a "normal-use scenario", keeping it open in a pinned tab.

Edit:
I use Ubuntu 16.10 on a Dell XPS 13 (9360)

offlinehoster:
Same problem with Ubuntu. High CPU usage by 105-110 %.

After closing that threema web app tab it is going down to the regulat 1-8 %....

Ubuntu 16.04 LTS (64bit)
Firefox 51.0.1 (64bit)

2b-as:
Same issue here on OSX 10.11.6 with Firefox 51.0.1, but _only_ in the messaging/chat view.

As soon as I return to the overview of contacs with their last messages, CPU usage drops to almost normal levels.

meyerjo:
Same issue on Windows 10 with 56.0.2924.87 (64-bit) (only in the messaging view)

rugk:
dupe of https://github.com/threema-ch/threema-web/issues/20 (or at least closely related)

fichtennadel:
Some details for this issue on FF 51.0.1 on Win7: 

The CPU load goes only up to 100%, when chat/messaging view is open. 
As soon as I close the indivdial chat by clicking ◀️ and returning to the contacts overview, CPU is down to idle again. 

Javascript debugger "pause" button stops repeatedly in `angular-material.min.js`:
`var n = e.Math.max(0, e.Math.min(i.now() - h, E));`

Call stack:

    f@https://web.threema.ch/node_modules/angular-material/angular-material.min.js?v=1.0.3:13822:0
    $@https://web.threema.ch/node_modules/angular-material/angular-material.min.js?v=1.0.3:13829:0
    callback@https://web.threema.ch/node_modules/angular/angular.js?v=1.0.3:12908:10
    completeOutstandingRequest@https://web.threema.ch/node_modules/angular/angular.js?v=1.0.3:6045:6
    Browser/self.defer/timeoutId<@https://web.threema.ch/node_modules/angular/angular.js?v=1.0.3:6324:6


Hope this helps.

dbrgn:
[...]

PR #264, opened 2017-07-10
Replace conversation loading CSS animation with GIF

dbrgn:
The animation runs a bit less smooth with the GIF but is more efficient on the CPU in less optimized browsers (like current Firefox stable).

The main problem was that some browsers – at least Firefox – process the animation frames of the CSS spinner even if it's currently invisible. Here's a profile from an already loaded, idle conversation:

img

Even though the spinner is invisible, the style is recalculated every few ms.

This is the version with the GIF, idle conversation:

img2

Only the garbage collection is active, no redrawing happens.

And here is the profile of a new conversation being loaded:

img3

The GIF animation is clearly visible (green "paint" events). But as soon as the messages have been loaded, it stops.

Refs #39


dbrgn:
Potentially related bugs:

- https://bugzilla.mozilla.org/show_bug.cgi?id=1166500 (should be fixed since FF 49)
- https://bugzilla.mozilla.org/show_bug.cgi?id=1237454 (not yet done)

PR #276, opened 2017-07-22
Make battery icon unselectable 

rugk:
Fixes threema-ch#275

It does allow to select the percentage of the battery though, as I think this is a legitimate thing to select. (When you want to share your battery stats… :laughing:)


dbrgn:
Thanks!

PR #278, opened 2017-07-26
Add/save contact/group/list using enter

### LINKED ISSUE #271: Make it possible to add new contact/group/list with 'Enter'
When you want to add another Threema contact it would be much more pleasant if you could do that with simple 'Enter'.
To touch the mouse just to click that tick on the right upper corner is not that nice. :smile: 

2017-07-17-195200_1366x768_scrot

Same, by the way, with creating a new group or distribution list.

## Your Environment

* Threema Web version: 1.3.1

IndianaDschones:
Fixes #271


dbrgn:
Thanks!

ovalseven8:
Well, it's not possible to use "Enter" for creating a group. At least when I've tested it. :confused:

dbrgn:
Yeah, because you need at least one member. But you can shift-tab back into the name and press enter there.

ovalseven8:
But when I already have one member and the cursor is at the "Choose members" field an Enter click does not help.

PR #282, opened 2017-08-02
Implement display of VoIP message types

dbrgn:
@sillych can you test?


src/directives/message_voip_status.ts
```diff
@@ -0,0 +1,64 @@
...
+            controllerAs: 'ctrl',
+            controller: [function() {
+                this.status = this.message.voip.status;
+                this.duration = this.message.voip.duration;
+                this.reason = this.message.voip.reason;
+                this.incoming = !this.message.isOutbox;
+                this.outgoing = this.message.isOutbox;
+            }],
+            template: `
+                <p ng-if="ctrl.status == 1">
```
IndianaDschones: I think it would be better to use strict equality `===` 😉

sillych: :+1:

dbrgn: indy

PR #287, opened 2017-08-05
Show peer connection type (direct or relayed)

### LINKED ISSUE #13: Show SaltyRTC connection status
User story: As a Threema Web user I want to see how my Threema App and Web client are connected.

This includes:
1. Either using a direct connection (LAN) or routing over the server (TURN).
2. Whether my mobile data is used (fees or so…) or whether my LAN/WLAN is used.

It would be good to combine these two similar (but not exactly the same) pieces of information into one status symbol or so…
Number 1 is what security geeks or people experiencing reconnections/performance issues would like to see and 2 is what people in special situations (limited data plan, …) or people interested in this generally like to see.

-----

Appended from https://github.com/threema-ch/threema-web/issues/195:

User story: I, as a (heavy) Threema Web user, want to see what route the Threema Web connection uses (mobile data/WiFi, uses TURN?) in order to evaluate problems/slowness/security/high costs…

This helps to:
* avoid high costs when users have limited mobile data plans
* evaluate the security (TURN is probably not what I want to use)
* find reasons for slowness ("ah, mobile data and TURN, …") or disconnections ("oh, just switched to TURN")
* potentially debug things (when you can also see what TURN mode is used – in a submenu, of course)

### Potential solution
Display an icon in the "Threema Web" panel in the Threema app showing a basic connection status. And maybe add a context item ("Connection details" or the like) where advanced users can see a more detailed status. (maybe even with connection duration or so)

lgrahl:
Changes:
* Add peer connection type directive, service and SASS
* The peer connection type is being refreshed every 10 seconds or when the ICE connection state changes
* Fix apply default cursor when selecting battery indicator, too
* Bump adapter JS version

Preview:
image

Resolves #13


lgrahl:
It probably doesn't solve #13 completely as there's no detailed info (but we could add a dialog for this purpose).

And I presume we will have a discussion about `srflx` candidates. :wink:

rugk:
As for the UI a small WLAN/wifi icon on the bottom right of the p2p icon could be displayed when the p2p connection is going through (W)LAN.

As for the logic: Is not it enough to know that a p2p connection is there and both clients use the same WLAN? Maybe exchange the wifi SSID to check that, or so…

lgrahl:
> Maybe exchange the wifi SSID to check that, or so…

In the browser?

You can probably get the information which interface the app is using for the connection but that's pretty much it. In reality, that should be sufficient because the mobile network is usually metered but for WiFi (and possibly even Bluetooth or USB) we usually don't care.

So yeah, mabe display a combination of both using the same colour scheme (amber for mobile, green for WiFI & co).

ovalseven8:
I don't know where but an explanation what the different icons with the different colors mean is necessary.

lgrahl:
Well, you hover them and it tells you whether it's direct or relayed.

dbrgn:
I'm not sure whether we should display this information in the bar directly. I think this is something that most people don't care about, or that will confuse them if they don't know much about how networks work.

I'd prefer an entry in the dropdown menu called "Network Information" or "Connection Information", where this information is displayed in a dialog in more detail. It could also contain a visualization, I already have a few ideas on how to display it graphically.

@sillych what's your opinion?

ovalseven8:
I really like the idea with the visualization. :+1:

lgrahl:
@dbrgn Something similar to this?

webrtc-data-paths-4

* *NAT* on either side would appear in case of `srflx`/`prflx` (even for NAT hairpinning, that's technically still correct). It would not appear for `relay` (technically, there could be a NAT in between but it doesn't matter).
* *Relay* would appear on either side in case of `relay`.
* Nothing would be in between in case of `host` on either side.

Be aware that this provides little or no information regarding whether a metered connection is being used. See [my earlier comment](https://github.com/threema-ch/threema-web/pull/287#issuecomment-320379792).

ovalseven8:
Just my two cents but I think it's too technical/detailed.

lgrahl:
Isn't that the whole point of a *Connection Information* dialog? But I certainly agree it should be stylistically improved with icons.

rugk:
> Edit: If there's a way to retrieve information which interface is being used for a specific IP address on Android, then that info could be sent to the web client and shown in that graph or as an additional warning icon as well.

Good idea: Can't you just check in the Android app whether wifi is used for the connection? I mean the other client does not matter then, because it must be in the wifi too. Or check in the app the IP addresses – if they belong to a local network…

dbrgn:
@lgrahl yeah, something like that, but with icons for the devices, servers and firewalls :) @sillych what do you think?

To make the implementation easier, the various configurations could be prepared as SVG images and switched out as applicable.

I agree that this dialog may be technical, but it should be "hidden" in the menu so that people that don't care about the connection aren't bothered about it.

sillych:
i like it...  :+1: 

use some cool material icons for the visualization (e.g. device_oder for the peers, screen_share for nat etc...)....what do you think @dbrgn / @lgrahl ?

lgrahl:
@dbrgn @sillych Feel free to nick my branch and add the changes for the dialog. No hard feelings on dropping my sloppy GUI proposal. :wink:

For the icons:
* smartphone for the app side
* public to illustrate a browser... maybe?
* I think dasboard would be a nice `srflx`/`prflx` icon... or router which will likely make more sense to non-techies.
* dns or cloud for `relay`.

N3dal:
@lgrahl For the browser icon you may use the icon which is used by Threema on the site for ThreemaWork. (https://work.threema.ch/)

For the devices/app you could also show the OS ( Android,...) running on the devices, even though it's not really necessary, because the user should know what his operating system is. But for support it could be helpful when somebody is posting a screenshot.

lgrahl:
I'm closing this one for now since it has been lingering a while and is heavily outdated.

PR #291, opened 2017-08-08
Show voip status messages in the navigation view

sillych:
@dbrgn please review


dbrgn:
Thanks!

ovalseven8:
Can someone provide a screenshot how it should look like?

Did a call today and nothing was displayed.

dbrgn:
This is about the messages regarding successful, missed and aborted calls in the past, not (yet) about live calls. Don't you see any of those messages?

ovalseven8:
Mh, I thought this commit is about something different, sorry. :)

PR #298, opened 2017-08-16
Replace npm-shrinkwrap.json with package-lock.json

dbrgn:
This is a newer dependency pinning mechanism by npm and superior for our use case.

This commit also updates some dependencies.


package-lock.json
```diff
@@ -0,0 +1,6667 @@
...
+          "integrity": "sha1-ZawFBLOVQXHYpklGsq48u4pfVPY=",
+          "dev": true,
+          "requires": {
+            "has-flag": "1.0.0"
+          }
+        }
+      }
+    },
+    "configstore": {
+
```
rugk: Hmm, SHA-1. Don't you have a SHA256 version?

PR #308, opened 2017-08-24
Downgrade angular-material

dbrgn:
The newer version (introduced in 69acfcc) would cause conversation / contact tabs in the navigation to disappear.

screenshot

The downgrade from 1.1.4 to 1.1.1 seems to fix that for now.


ovalseven8:
:confused:

PR #316, opened 2017-08-31
Fix #305 and removed absolute (max-)heights and gave the header (with its logo) more space.

### LINKED ISSUE #305: Focus should be in input after sending message
## Expected Behavior

After sending a message the focus should **always** be on the input field, so I can just continue to type the next message.

## Current Behavior

When clicking on the sent button, the focus seems to stay on that button.

1. Type a message.
2. Click on the "sent button".

JlnWntr:
A little design quillet.

### DIFF
--- a/src/directives/compose_area.ts
+++ b/src/directives/compose_area.ts
@@ -193,6 +193,7 @@ export default [
                         submitText().then(() => {
                             // Clear compose div
                             composeDiv[0].innerText = '';
+                            composeDiv[0].focus();
 
                             // Send stopTyping event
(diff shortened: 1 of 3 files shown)


dbrgn:
Hello and thanks for your pull request!

I merged the fix for the input field focus in f710a5b.

We decided to keep the tighter spacing in the header, we prefer to see everything on the screen at a glance without scrolling but on the other hand do not want to waste too much screen space in the header.

JlnWntr:
Okay, cool.

PR #323, opened 2017-09-07
Show group creation and member count in group detail view

### LINKED ISSUE #255: Group details improvements
- Add group creator back to group members for groups where current user is the creator
- List number of members in group and group member limit
- List group creation date

sillych:
Similar to the android app group view (Group creation date value available after the next app update)

### DIFF
--- a/src/threema.d.ts
+++ b/src/threema.d.ts
@@ -245,6 +245,7 @@ declare namespace threema {
         members: string[];
         administrator: string;
         access: GroupReceiverAccess;
+        createdAt: string;
     }
 
     /**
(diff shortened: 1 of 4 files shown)


sillych:
group-1

src/threema.d.ts
```diff
@@ -245,6 +245,7 @@ declare namespace threema {
         members: string[];
         administrator: string;
         access: GroupReceiverAccess;
+        createdAt: string;
```
dbrgn: This should be optional (`createdAt?`), since older apps don't send it, right?

sillych: yes!

sillych:
Fix #255

rugk:
In the screenshot, I cannot see the group creation date requested in https://github.com/threema-ch/threema-web/issues/255.

dbrgn:
@rugk because the group was created the same day.

sillych:
@rugk he same behaviour like threema on android.

PR #329, opened 2017-09-08
implement clean method delete all message of a chat

### LINKED ISSUE #50: Delete whole chat
At the moment, it is possible to delete single messages with Threema Web.

The feature to delete a whole chat would be cool too.

sillych:
Call "Delete Chat" in the receiver detail view to clean a chat (similar to the android app).

2017-09-08-132131_1501x694_scrot

Requires a app update

Fix #50

### DIFF
--- a/public/i18n/de.json
+++ b/public/i18n/de.json
@@ -151,7 +151,9 @@
         "UNKNOWN_MESSAGE_TYPE": "Unbekannter Nachrichtentyp",
         "NICKNAME": "Nickname",
         "THREEMA_WORK_CONTACT": "Threema Work Nutzer",
-        "THREEMA_BLOCKED_RECEIVER": "blockiert"
+        "THREEMA_BLOCKED_RECEIVER": "blockiert",
+        "DELETE_THREAD": "Chat löschen",
+        "DELETE_THREAD_MESSAGE": "{count, plural, one {Möchten Sie wirklich diesen Chat löschen? Die Nachrichten können nicht wiederhergestellt werden.} other {Möchten Sie wirklich # Chats löschen? Die Nachrichten können nicht wiederhergestellt werden.}}"
     },
     "messageTypes": {
         "AUDIO_MESSAGE": "Sprachnachricht",
--- a/src/controller_model/contact.ts
+++ b/src/controller_model/contact.ts
@@ -113,8 +113,6 @@ export class ContactControllerModel implements threema.ControllerModel {
 
     public canClean(): boolean {
         return this.canView();
-            // blocked
-            // && this.contact.;
     }
 
     public clean(ev: any): any {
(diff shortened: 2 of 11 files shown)


src/controller_model/contact.ts
```diff
@@ -111,6 +111,43 @@ export class ContactControllerModel implements threema.ControllerModel {
             );
     }
 
+    public canClean(): boolean {
+        return this.canView();
+            // blocked
+            // && this.contact.;
```
dbrgn: Dead code

public/i18n/de.json
```diff
@@ -150,7 +150,9 @@
         "BALLOT_MESSAGES_NOT_SUPPORTED": "Umfragen werden in Threema Web derzeit nicht untersützt.",
         "UNKNOWN_MESSAGE_TYPE": "Unbekannter Nachrichtentyp",
         "NICKNAME": "Nickname",
-        "THREEMA_WORK_CONTACT": "Threema Work Nutzer"
+        "THREEMA_WORK_CONTACT": "Threema Work Nutzer",
+        "DELETE_THREAD": "Chat löschen",
+        "DELETE_THREAD_MESSAGE": "{count, plural, one {Möchten Sie wirklich diesen Chat löschen? Die Nachrichten können nicht wiederhergestellt werden.} other {Möchten Sie wirklich # Chats löschen? Die Nachrichten können nicht wiederhergestellt werden.}}"
```
dbrgn: Do you plan to support deleting multiple chats at once? So far we only use count 1. If not, I'd remove the plural forms.

src/controller_model/contact.ts
```diff
@@ -111,6 +111,43 @@ export class ContactControllerModel implements threema.ControllerModel {
...
+    }
+
+    private reallyClean(): any {
+        if (!this.canClean()) {
+            this.$log.error('not allowed to clean this contact');
+            return;
+        }
+
+        this.isLoading = true;
+        this.webClientService.clean(this.contact)
```
dbrgn: Maybe the `clean` method should be renamed to `cleanChat` or `deleteChat` to make it easier to understand what this does.

src/services/webclient.ts
```diff
@@ -1227,6 +1228,24 @@ export class WebClientService {
     }
 
     /**
+     * Remove all messages of a receiver
+     * @param {threema.Receiver} receiver
+     * @returns {Promise<any>}
+     */
+    public clean(receiver: threema.Receiver): Promise<any> {
+        if (receiver === undefined) {
```
dbrgn: We should probably check for `null` too. (The joys of Javascript.)

sillych: feel free to append this check :laughing:

public/i18n/de.json
sillych: yes is planned...

sillych:
- If the App does not support this feature, nothing happens, i think its ok.
- After cleaning a conversation, the cache (stored messages on browser) will be cleared now
- Chats will be closed now (app) after a clean (webclient)

PR #336, opened 2017-09-11
show caption input only on file and image messages

### LINKED ISSUE #332: Video description not displayed when I send video
When I send a video (not as file!) in Threema Web, I can write a description.

However, this description is not displayed!

In the app it's not possible to write a description if I send a video as a video.

2017-09-09-113024_1920x1080_scrot

2017-09-09-113109_1920x1080_scrot

sillych:
Show caption input field only if the message should send as file or the media is a image.
The app only support captions in file and image messages (not in video or audio messages).
Fix #332

### DIFF
--- a/src/partials/messenger.ts
+++ b/src/partials/messenger.ts
@@ -427,6 +427,7 @@ class ConversationController {
                 case 'file':
                     // Determine file type
                     let showSendAsFileCheckbox = false;
+                    let captionSupported = false;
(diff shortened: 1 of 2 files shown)


dbrgn:
Cool! :) But now the caption box is not shown for plain files (e.g. a PDF). Maybe the default of the variable should be `true`?

sillych:
done!

PR #338, opened 2017-09-11
emojify nicknames

sillych:
use emojify parser in

my nickname:
from
2017-09-11-134648_336x110_scrot
to
2017-09-11-135305_266x106_scrot

and in the contact detail view
from
2017-09-11-135540_854x249_scrot
to
2017-09-11-135710_834x278_scrot

### DIFF
--- a/src/directives/my_identity.ts
+++ b/src/directives/my_identity.ts
@@ -55,9 +55,7 @@ export default [
             }],
             template: `
                 <div class="my-threema-information" ng-click="ctrl.showQRCode()">
-                    <div class="threema-id" ng-cloak>
-                        <span translate>messenger.MY_PUBLIC_NICKNAME</span>
-                        {{ctrl.identity.publicNickname}}
+                    <div class="nickname" ng-cloak ng-bind-html="ctrl.identity.publicNickname | emojify">
                     </div>
                 </div>
             `,
--- a/src/directives/my_identity.ts
+++ b/src/directives/my_identity.ts
@@ -55,7 +55,8 @@ export default [
             }],
             template: `
                 <div class="my-threema-information" ng-click="ctrl.showQRCode()">
-                    <div class="nickname" ng-cloak ng-bind-html="ctrl.identity.publicNickname | emojify">
+                    <div class="nickname" ng-cloak
+                        ng-bind-html="ctrl.identity.publicNickname | emojify | emptyToPlaceholder: '-'">
                     </div>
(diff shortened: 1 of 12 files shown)


dbrgn:
Now the Identity is not shown anymore if there is no nickname defined :)

sillych:
That’s old hat! Check again

src/directives/my_identity.ts
```diff
@@ -55,7 +55,8 @@ export default [
             }],
             template: `
                 <div class="my-threema-information" ng-click="ctrl.showQRCode()">
-                    <div class="nickname" ng-cloak ng-bind-html="ctrl.identity.publicNickname | emojify">
+                    <div class="nickname" ng-cloak
+                        ng-bind-html="ctrl.identity.publicNickname | emojify | emptyToPlaceholder: '-'">
```
dbrgn: At first I thought this is an exploit (if the nickname contains a `<script>` tag), but Angular sanitizes input in ng-bind-html b default. :+1:

src/sass/sections/_conversation.scss
```diff
@@ -301,6 +300,11 @@
             font-weight: bold;
             font-size: 0.9em;
             margin-bottom: 8px;
+            // Emoji styling
+            > span:not(.ng-binding) {
+                transform: scale(0.6);
+                //top: 2px
```
dbrgn: This can be removed.

src/directives/my_identity.ts
```diff
@@ -55,9 +55,8 @@ export default [
             }],
             template: `
                 <div class="my-threema-information" ng-click="ctrl.showQRCode()">
-                    <div class="threema-id" ng-cloak>
-                        <span translate>messenger.MY_PUBLIC_NICKNAME</span>
-                        {{ctrl.identity.publicNickname}}
+                    <div class="nickname" ng-cloak
+                        ng-bind-html="ctrl.identity.publicNickname | emojify | emptyToPlaceholder: '-'">
```
dbrgn: Shouldn't we use the identity as fallback, instead of `-`?

src/sass/sections/_conversation.scss
sillych: feel free :-)

src/directives/my_identity.ts
sillych: no, it should be not possible to set a blank nickname in the app :-)

PR #339, opened 2017-09-12
Ask the user to update the app if reconnecting fails

dbrgn:
This warning is shown if `PREV_PROTOCOL_LAST_VERSION` in `config.ts` is
set. That config variable should be set in case a new protocol version
was released for Threema Web.

English

German

The link points to the configured version in the [Threema Web Archive](https://web.threema.ch/archive/).

### DIFF
--- a/README.md
+++ b/README.md
@@ -69,6 +69,10 @@ The configuration of Threema Web can be tweaked in `src/config.ts`:
 
 - `SELF_HOSTED`: Set this to `true` if this instance of Threema Web isn't being
   hosted on `web.threema.ch`.
+- `PREV_PROTOCOL_LAST_VERSION`: When the Threema Web protocol version changes,
+  this can be set to the last version of Threema Web that supported  the
+  previous protocol version. If set to something different than `null`, a
+  message will be shown to the user if reconnecting fails.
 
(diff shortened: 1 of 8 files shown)


ovalseven8:
Will Threema Web use a new protocol soon? If yes, can you explain the reasons/advantages? :wink: 

One note:
I am not sure if it's good practice to link to old Threema web versions with more bugs and less features. Is not forcing the users to install the latest app version the best and only thing to consider?

dbrgn:
> Will Threema Web use a new protocol soon? If yes, can you explain the reasons/advantages?

Yes, containing small, reverse-incompatible changes, or general protocol cleanup. Things like [this](https://github.com/threema-ch/threema-web/pull/254).

> Is not forcing the users to install the latest app version the best and only thing to consider?

The best thing, yes. But it's a good temporary solution if people can't or don't want to upgrade for some reason. Or if we're doing a staged rollout and not all people have received the new version yet.

rugk:
In any case, when vulnerabilities are fixed in Threema Web, you should of course pull down previous versions with that vulnerability. Because considering thee might be an XSS vuln. that would otherwise stay on the site…

PR #340, opened 2017-09-13
Battery status improvements

### LINKED ISSUE #268: Show warning when the battery is low
When using WhatsApp Web and the battery is low, a warning is displayed:
2017-07-16-213015_1366x768_scrot

Would make sense for Threema Web too!

### LINKED ISSUE #288: No translation
My default language is German, but it's not translated.
Moreover: Why is not my current battery displayed when below 20 % (only hover)?

2017-08-06-194932_1366x768_scrot

### LINKED ISSUE #309: Battery level hidden when charging
## Expected Behavior

Show battery level, always. Just because you charge your device, you still want to see the current battery level (or want to see how it rises :wink:).

## Current Behavior

The battery level is hidden when the phone is charged. I can only hover over the icon to see the percentage.

1. Charge your Android device.

## Your Environment

* Threema Web version: 1.4.1-gh

sillych:
Always show percent (fix #309) view
2017-09-13-133304_112x63_scrot

 translate charging/discharging text (fix #288) and blink percent view on alert (fix #268)
2017-09-13-133350_93x51_scrot

### DIFF
--- a/public/i18n/de.json
+++ b/public/i18n/de.json
@@ -238,6 +238,7 @@
     },
     "battery": {
         "CHARGING": "Aufladen: {percent}%",
-        "DISCHARGING": "Entladen: {percent}%"
+        "DISCHARGING": "Entladen: {percent}%",
+        "ALERT": "Entladen: {percent}%"
     }
 }
--- a/public/i18n/de.json
+++ b/public/i18n/de.json
@@ -235,6 +235,9 @@
         "CALL_FINISHED_OUT": "Ausgehender Anruf",
         "CALL_REJECTED": "Anruf abgelehnt",
         "CALL_ABORTED": "Anruf abgebrochen"
+    },
+    "battery": {
+        "CHARGING": "Aufladen: {percent}%",
+        "DISCHARGING": "Entladen: {percent}%"
     }
-
 }
--- a/public/i18n/en.json
+++ b/public/i18n/en.json
@@ -240,6 +240,7 @@
     },
     "battery": {
         "CHARGING": "Charging: {percent}%",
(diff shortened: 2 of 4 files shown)


ovalseven8:
The red font is hard to read. :/

sillych:
Thanks to @ovalseven8, alert text back to white
2017-09-13-141811_54x63_scrot

ovalseven8:
:+1: 

I am not sure if the animation makes sense though. Looking for alternatives at the moment, mh...

rugk:
The icon (the outline, at least) could be made red. And IMHO this does not yet fix https://github.com/threema-ch/threema-web/issues/268, because such a small icon is not enough indication.

As shown in the issue https://github.com/threema-ch/threema-web/issues/268, you should probably use a banner, i.e. something big with orange background or so, and, you may also show a desktop notification, as otherwise, when the user is just running Threema Web in the background and no messages arrive, it will not notice the battery discharge at all. Only when the user later discovers the connection is lost and the phone is off… but then, this is too late.

ovalseven8:
Well, I would keep the icon plus font white and remove the animation.

And, yeah, the banner is the best option probably.

dbrgn:
The entire icon blinks now.

> And IMHO this does not yet fix #268

We feel it does. It's less intrusive than a big banner and takes away less screen space. But it still calls for attention.

rugk:
Okay, that's good if it is not too 80th-style… :wink: 

I'd say a banner is less intrusive (just make it dismissable, i.e. let the user close/hide it). If I'd want to ignore the blinking icon, that is really hard… but when it is only for < 10% it may be okay…

What about a desktop notification?

PR #341, opened 2017-09-13
Update STUN/TURN endpoints

dbrgn:


lgrahl:
DS for dual stack?

dbrgn:
> DS for dual stack?

Yes. (The "naked" previously used subdomains also had dual stack enabled until today, so this commit doesn't change anything.)

PR #343, opened 2017-09-13
#318|#319

### LINKED ISSUE #318: No character limit for image description
When I write a normal message there is a character limit (maximally 3500 chars per message). However, when I send an image, the image description can contain as many chars as you like.

Perhaps needs also app update, don't know.

### LINKED ISSUE #319: Image description not readable if it requires more than one line
If the image description requires more than one line in the image view in Threema Web, it's not fully readable.

2017-09-02-225226_1920x1080_scrot

sillych:
- limit caption text to 1000 characters
- show only one line of the caption text in the mediabox

fix #318
fix #319


dbrgn:
:ok_hand:

PR #351, opened 2017-09-27
Show desktop notifications if battery level is low

### LINKED ISSUE #348: Show notification when battery is (nearly) dead
Followup of https://github.com/threema-ch/threema-web/issues/268 / https://github.com/threema-ch/threema-web/pull/340

**Reason:** I just again had the issue that I was chatting, got connection issues, (fortunately copied my text out of Threema Web), the connection was lost and then I saw that my battery was dead.

The thing is, you just do not notice [a tiny icon/text](https://github.com/threema-ch/threema-web/pull/340#issue-257349055) there. It needs to make you aware that you really have a problem if you do want to continue to use Threema Web. It needs a **warning**.
You also don't notice the indicators of the phone itself, as it may be laying somewhere else and you are less likely to take it up if you are already using Threema Web.

IMHO the best implementation would be:
* a browser notification, as the user will see this, even when not using the browser/doing something else

IMHO the notification should be triggered:
* at 15%, "a serious warning"
* at 5% when it is nearly dead

dbrgn:
There are two different levels: Low (<20%) and critical (<5%).

If desktop notifications are disabled by the user, no notification is triggered.

Chrome, low, DE:

screenshot1

Firefox, critical, EN:

screenshot2

Fixes #348


N3dal:
How about replacing the german word "tief" with "niedrig". I think the sentence sounds better with this word.

dbrgn:
> I think the sentence sounds better with this word.

Point taken, that was probably a helvetism :)

PR #356, opened 2017-10-10
Fix centering of caption text

### LINKED ISSUE #352: Position of image caption sometimes wrong
## Expected Behavior

Caption centered below image.

## Current Behavior

Especially with portrait photos the caption is often at the left instead of centered at the middle

position

## Your Environment

* Threema Web version: 1.6.0-gh (I think the issue was introduced in this version.)
* Browser name and version: Firefox 55

dbrgn:
Fixes #352


sillych:
:taco:

PR #360, opened 2017-10-18
Don't display low battery notification when charging

### LINKED ISSUE #359: Don't display empty battery notification when charging
## Expected Behavior
A push notification for empty battery should only be displayed when the battery is empty AND the device is not charging.

## Current Behavior
A push notification for empty battery is displayed if the battery is empty no matter if it is plugged or not.

1. Use your phone until its battery is below 19%
2. Plug your phone to the charger and open Threema Web

## Your Environment

* Threema Web version: 1.6.1
* Threema for Android version: 3.21.2000380
* Browser name and version: Chrome 62
* Operating system and version: Ubuntu 17.04

dbrgn:
Fixes #359.

### DIFF
--- a/src/services/battery.ts
+++ b/src/services/battery.ts
@@ -45,12 +45,14 @@ export class BatteryStatusService {
         this.batteryStatus = batteryStatus;
 
         // Alert if percent drops below a certain threshold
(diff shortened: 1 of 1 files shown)


Elsensee:
Wow, that was quick! :) I personally would also hide the notification when the phone is charging but I don't think that it really matters much.

PR #363, opened 2017-10-25
Link to changelog from version update alert dialog

dbrgn:
When notifying the user about a new version of Threema Web, link to the changelog.

screenshot


ovalseven8:
What about
> Click "OK" to activate the update

dbrgn:
That's actually a good suggestion :)

PR #368, opened 2017-11-02
Switch to CircleCI 2.0

dbrgn:
Implement a real test matrix.


dbrgn:
I can't get the build step to run successfully on Node 4 or 6 in CI. Will have to upgrade the babel toolchain anyways, maybe that will fix the issue.

PR #369, opened 2017-11-02
Warn if connecting to an already connected session

dbrgn:
This shows a warning when trying to connect to a stored session that is already connected in another tab or window.

screenshot

Thanks @lgrahl for the initial draft and for bringing the [BroadcastChannel API](https://developer.mozilla.org/en-US/docs/Web/API/BroadcastChannel) to my attention!

### DIFF
--- a/public/i18n/de.json
+++ b/public/i18n/de.json
@@ -34,7 +34,9 @@
         "MANUAL_START_STEP2": "Tippen Sie auf die Sitzung, die Sie wiederherstellen möchten",
         "MANUAL_START_STEP3": "Wählen Sie im Menu \"Sitzung starten\"",
         "MORE_ABOUT_WEB": "Mehr über Threema Web",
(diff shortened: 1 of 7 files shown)


dbrgn:
@lgrahl feel free to test this too, if you want!

lgrahl:
Beautiful! But... I can't get it to work with `npm run serve:live`.

dbrgn:
@lgrahl so what's the problem?

lgrahl:
1. I open a tabs for `localhost:9966` in Opera.
2. Create a new permanent Threema Web session.
3. Unlock that session in another tab for `localhost:9966`.
4. I don't get the screen shown above. It just behaves as before.

dbrgn:
Hm, it seems to fail if the session was newly created, but it works if the session is already restored in the first tab.

I'll have to take a closer look at this later :)

dbrgn:
@lgrahl should be fixed now, can you test?

lgrahl:
Yeah, that works for me now. Maybe it makes sense to add an additional check for a temporary session if another session is being created? Only if that's not too overkill. :wink:

dbrgn:
Other temporary sessions aren't really a problem, right? Since they use different keys, they're completely independent.

lgrahl:
Yeah, my idea would require a device check. Never mind.

PR #371, opened 2017-11-07
Use powershell to build the "dist" package on Windows

Octoate:
Hi,
I started to work with the repository on my Windows system today and noticed that it was not possible to build the web package with "npm run dist", because the script were not able to invoke the "package.sh" in the "dist" folder. However, if I invoked it manually from my Powershell window, it is working fine (it invokes the bash in the Ubuntu subsystem that I have installed on my system).
To build the distribution package on my system I added the "build-package.js" in the "dist" directory that decides how the "package.sh" script shall be invoked on the different systems and added "powershell" to the command invokation on Windows systems. The "dist" script in the "package.json" was modified to invoke the "build-package.js" script with Node.js. Now it should build fine on Windows systems, too.

CU Tim

### DIFF
--- a/dist/build-package.js
+++ b/dist/build-package.js
@@ -1,15 +1,12 @@
-var exec = require('child_process').exec;
-function puts(error, stdout, stderr) { console.log(stdout) }
-
+var spawn = require('child_process').spawn;
 var os = require('os');
-//control OS
-//then run command depengin on the OS
 
-if (os.type() === 'Linux') 
-   exec("dist/package.sh", puts); 
-else if (os.type() === 'Darwin') 
-   exec("dist/package.sh", puts); 
-else if (os.type() === 'Windows_NT') 
-   exec("powershell dist/package.sh", puts);
-else
+if (os.type() === 'Linux') {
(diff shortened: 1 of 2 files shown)


dbrgn:
Thanks for this pull request, @Octoate!

On Linux, the build (triggered with `node dist/build-package.js`) seems to work the first time, but the second time the prompt just hangs without any output.

I think the reason is that node somehow buffers the output and does not properly handle the input that is required when prompting whether the "release" directory should be deleted (in `package.sh`).

Can you check whether it's possible to fix that?

Octoate:
Hi @dbrgn,
I will have a look at it as soon as I have some time left and a VM with Linux up and running. Hopefully I am able to find the issue with it.

dbrgn:
@Octoate I pushed a change to your branch, that seems to do the trick on Linux. Could you test whether it works on Windows too?

Octoate:
Thanks that you investigated the issue under Linux. I just checked it under Windows and it works fine and creates the distribution package as expected. Interesting that using "spawn" instead of "exec" fixed it.

dbrgn:
Thanks! Hope everything works now as intended 🙂

PR #404, opened 2018-01-26
Mention

sillych:
Support the new mention feature (Display Only)

2018-01-26-142618_832x268_scrot

will be

2018-01-26-142645_805x299_scrot

### DIFF
--- a/src/filters.ts
+++ b/src/filters.ts
@@ -15,7 +15,7 @@
  * along with Threema Web. If not, see <http://www.gnu.org/licenses/>.
  */
 
-import {filter} from './helpers';
+import {escapeRegExp, filter} from './helpers';
 import {MimeService} from './services/mime';
 import {WebClientService} from './services/webclient';
 
@@ -152,8 +152,9 @@ angular.module('3ema.filters', [])
 
                         if (html !== undefined) {
                             text = text.replace(
-                                new RegExp(possibleMention.replace(/([.*+?^=!:${}()|\[\]\/\\])/g, '\\$1'), 'g'),
-                                html);
+                                new RegExp(escapeRegExp(possibleMention), 'g'),
+                                html,
+                            );
                         }
                     }
                 }
--- a/src/filters.ts
+++ b/src/filters.ts
@@ -127,6 +127,41 @@ angular.module('3ema.filters', [])
 })
 
 /**
+ * Convert mention elements to html elements
+ */
+.filter('mentionify', ['WebClientService', '$translate',
+    function (webClientService: WebClientService, $translate) {
+        return(text) => {
+            if (text !== null && text.length > 10) {
+                let result = text.match(/@\[([\*\@a-zA-Z0-9][\@a-zA-Z0-9]{7})\]/g);
+                if (result !== null) {
+                    result = ([...new Set(result)]);
+                    // Unique
+                    for (let possibleMention of result) {
+                        let identity = possibleMention.substr(2, 8);
+                        let html;
+
+                        if (identity === '@@@@@@@@') {
+                            html = '<span class="mention all">' + $translate.instant('messenger.ALL') + '</span>';
+                        } else {
+                            const contact = webClientService.contacts.get(possibleMention.substr(2, 8));
+                            if (contact !== null) {
+                                html = '<span class="mention contact">' + contact.displayName + '</span>';
+                            }
+                        }
+
+                        if (html !== undefined) {
+                            text = text.replace(
+                                new RegExp(possibleMention.replace(/([.*+?^=!:${}()|\[\]\/\\])/g, '\\$1'), 'g'),
+                                html);
+                        }
+                    }
+                }
+            }
+            return text;
+        };
+}])
+/**
  * Reverse an array.
  */
 .filter('reverse', function() {
--- a/tests/filters.js
+++ b/tests/filters.js
@@ -4,17 +4,49 @@ describe('Filters', function() {
 
     // Ignoring page reload request
     beforeAll(() => window.onbeforeunload = () => null);
+    let webClientServiceMock = {
+        contacts: {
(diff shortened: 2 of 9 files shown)


src/filters.ts
```diff
@@ -126,6 +126,37 @@ angular.module('3ema.filters', [])
...
+            if (text !== null && text.length > 10) {
+                let result = text.match(/@\[([\*\@a-zA-Z0-9][\@a-zA-Z0-9]{7})\]/g);
+                if (result !== null) {
+                    result = ([...new Set(result)]);
+                    // Unique
+                    for (let possibleMention of result) {
+                        let identity = possibleMention.substr(2, 8);
+                        if (identity === '@@@@@@@@') {
+                            text = text.replace(
+                                new RegExp(possibleMention.replace(/([.*+?^=!:${}()|\[\]\/\\])/g, '\\$1'), 'g'),
```
dbrgn: Whoa, I don't understand this RegEx. What does it do?

Couldn't we simply do `text.replace('@[@@@@@@@@]', <span class="mention all">' + $translate.instant('messenger.ALL') + '</span>')` first, and only then search for contact mentions?

tests/filters.js
```diff
@@ -121,4 +153,50 @@ describe('Filters', function() {
...
+                ['@[AAAAAAAA]', '<span class="mention contact">ContactA</span>'],
+                ['hello @[AAAAAAAA]. @[AAAAAAAA] you are my friend', 'hello <span class="mention contact">ContactA</span>. <span class="mention contact">ContactA</span> you are my friend'],
+                ['@[AAAAAAAA] @[AAAAAAAA] @[AAAAAAAA]', '<span class="mention contact">ContactA</span> <span class="mention contact">ContactA</span> <span class="mention contact">ContactA</span>']
+            ]);
+        });
+
+        it('mention - all', () => {
+            this.testPatterns([
+                ['@[@@@@@@@@]', '<span class="mention all">messenger.ALL</span>'],
+                ['@[@@@@@@@@] your base are belong to us', '<span class="mention all">messenger.ALL</span> your base are belong to us'],
```
dbrgn: I approve of this test case!

sillych: 👍

src/filters.ts
sillych: this saves one iteration, replace will be only called if one or more mentions (including the `all` mention) are found

sillych: check my new commit

sillych:
padding changed to `2px 5px;` 👾

dbrgn:
When having mentions on two lines directly above each other, they partially overlap. But that's not too terrible, so it can still be fixed later :)

rugk:
IMHO you should always still display the `@` char at the beginning. Also in the app, if it is not shown there.
Just so users know that this is a mention and see they can do the same by entering that character. Just how every other char app does it, so…

ovalseven8:
> When having mentions on two lines directly above each other, they partially overlap. But that's not too terrible, so it can still be fixed later :)

Better create an issue :D

mimiks:
> you should always still display the @ char at the beginning
The @ char looks pretty awkward. IMHO the grey frame is prominent enough.

joelfischerr:
But the @ character is pretty much the universal character used for mentioning someone. I agree with @rugk on showing the '@' char at the beginning of every mention.

mimiks:
It might indeed be helpful to new users. But on the other hand, displaying the @ character at the beginning of a mention is redundant and doesn't look good.

rugk:
It is done like that in every chat application I know. So it confuses me a bit that this is not shown, even if I am not a beginner. 😄 
Also on GitHub it looks good this way, and seriously, no, there is no visual problem with an @ character. Or do you also don't like $ or # or what?

mimiks:
> It is done like that in every chat application I know

Don't think so. Check out Telegram for instance. And even if it was, we don't have to do it same way.

joelfischerr:
No, it's not redundant in my opinion as it clearly shows that someone is being mentioned. Just the highlighting of the text could lead to confusion as all names might be highlighted or it could be some formatting option. And as you can see in my reply above github shows the '@' character as well. I think it'd be better if it's done like everywhere else.

lgrahl:
`@`, `#` & co don't work for me here because real names contain spaces and that looks awkward (and FWIW would be awkward to parse, too).

rugk:
In MyBB it also allows spaces... But you already did the parsing work, so there is nothing to worry, it is just a visual change (i.e., it can even be accomplished with a CSS pseudo-class `::before`). (And I think the Threema IDs are used internally for the text, so it is not even a problem for the parser.)
It may look a bit surprising, but considering that you already have this grey background one clearly sees where the name starts and ends, i.e. it is just another character at the beginning, which is not a problem IMHO.

Maybe Threema can do an A/B test... 😄

lgrahl:
Right, Threema IDs are used internally. The whole point of having an `@` in front of it (for me) is that I can copy and paste that text and it does the same thing - mention a person. Which it wouldn't in this case, so it makes no sense.

joelfischerr:
It wouldn't preserve the mention when the text is copied back into Threema or when the text is copied elsewhere?
Because with the `@`in front of a name the meaning would be preserved in both cases. Without the `@`the meaning gets lost as soon as the text is shared.

> @, # & co don't work for me here because real names contain spaces and that looks awkward (and FWIW would be awkward to parse, too).

I don't think it looks awkward. The highlighting makes the block `@Max Musterman` look like one single entity.

rugk:
> Because with the @in front of a name the meaning would be preserved in both cases. 

No, because internally it is `@ZU732H` or so, with the Threema ID. Threema just shows the name instead of the ID.
So either it needs to parse the name, too, in `@` statements (and here we get to the spaces problem), or you just cannot copy the element.

PR #406, opened 2018-01-26
Fixes #405: Building Threema Web on macOS fails 

### LINKED ISSUE #405: Building Threema Web on macOS fails
## Expected Behavior
The self_hosting.md should allow building Threema Web in macOS.

## Current Behavior
macOS and Linux have some slight differences in how things work in bash. These differences make it impossible to build Threema Web on macOS.

## Possible Solution
Change the package.sh to allow it to distinguish between macOS and Linux and change the following things:
- Use `ditto` instead of `install -D` because `install -D` doesn’t work on macOS
- On macOS the version number could sometimes contain a single leading whitespace, add another sed to remove all whitespaces in the version number
- `sed` does not behave in the same way on macOS as it does on Linux, change `sed` to `gnu-sed` when adding the version number (`gnu-sed` can be installed from homebrew)
- Add Note about how to build on macOS to self_hosting.md
- Misc. add `.DS_Store` to `.gitignore`

1. call `npm run build` on macOS, it will fail.

## Your Environment

* Threema Web version: 1.7.0
* Operating system and version: macOS 10.13.3 High Sierra

joelfischerr:
As described in #405 there are some minor changes between bash on Linux and macOS. I think it would be nice if Threema Web could also be built on macOS.
These additions should allow building on macOS to succeed without breaking building on Linux.


joelfischerr:
With the additional commits from 2 days ago everything should work and produce the same output as it does on Linux.
I'll read and sign the CLA within the next week. Should I close the pull request in the meantime or is it fine if it's kept open?

rugk:
> Should I close the pull request in the meantime or is it fine if it's kept open?

Keep it open, the bot will detect when you sign the CLA.

dist/package.sh
```diff
@@ -78,12 +78,23 @@ targets=(
...
 
+
 for target in "${targets[@]}"; do
-    install -D "node_modules/$target" "$DIR/node_modules/$target"
+
+	if [[ "$OSTYPE" == "darwin"* ]]; then
+		ditto "node_modules/$target" "$DIR/node_modules/$target"
+    else
+		install -D "node_modules/$target" "$DIR/node_modules/$target"
+    fi
```
dbrgn: You have a mixture of spaces and tabs in your files. Can you fix that? :) We use 4 spaces to indent.

dist/package.sh
```diff
@@ -78,12 +78,23 @@ targets=(
...
 done
 
+
 echo "+ Update version number..."
-sed -i "s/\[\[VERSION\]\]/${VERSION}/g" $DIR/index.html $DIR/troubleshoot/index.html $DIR/dist/app.js $DIR/version.txt
+if [[ "$OSTYPE" == "darwin"* ]]; then
+	    gsed -i "s/\[\[VERSION\]\]/${VERSION}/g" $DIR/index.html $DIR/troubleshoot/index.html $DIR/dist/app.js $DIR/version.txt
```
dbrgn: What's the problem with regular `sed`? I don't think the user should have to install any additional build dependencies to run the build script.

joelfischerr: There are some differences between sed on versions of Unix using the GNU tools and sed on versions based on freeBSD like macOS (https://unix.stackexchange.com/questions/13711/differences-between-sed-on-mac-osx-and-other-standard-sed).
I'm not sure what needs to be changed to get it running on both platforms and thought that installing gsed is only a minor inconvenience.

dist/package.sh
```diff
@@ -28,7 +28,7 @@ if [ -e "release" ]; then
     done
 fi
 
-VERSION=$(grep "\"version\"" package.json  | sed 's/\s*\"version\": \"\([^\"]*\).*/\1/')
+VERSION=$(grep "\"version\"" package.json  | sed 's/\s*\"version\": \"\([^\"]*\).*/\1/' | sed 's/ //g')
```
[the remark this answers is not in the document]
joelfischerr: This is most likely also an issue with sed on macOS just like in line 86 https://github.com/threema-ch/threema-web/pull/406#discussion_r166021434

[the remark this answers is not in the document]
joelfischerr: I have changed this to also use gsed on macOS. But I'll look into what can be done to make it work on both platforms.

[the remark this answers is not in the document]
dbrgn: Apparently the sed on macOS does not support the `\s` character class properly. Can you try replacing `\s` with `[[:blank:]]`? That should do the trick even with regular `sed`.

dist/package.sh
dbrgn: This should work with regular `sed` on macOS if you use `sed -i ''` instead of `sed -i`.

dist/package.sh
```diff
@@ -28,7 +28,12 @@ if [ -e "release" ]; then
     done
 fi
 
-VERSION=$(grep "\"version\"" package.json  | sed 's/\s*\"version\": \"\([^\"]*\).*/\1/')
+if [[ "$OSTYPE" == "darwin"* ]]; then
+        VERSION=$(grep "\"version\"" package.json  | sed 's/[[:blank:]]*\"version\": \"\([^\"]*\).*/\1/')
+    else
+       VERSION=$(grep "\"version\"" package.json  | sed 's/\s*\"version\": \"\([^\"]*\).*/\1/')
+    fi
```
dbrgn: Sorry if my last comment was confusing. You can use `[[:blank:]]` for both darwin and non-darwin, so the `if` can be removed altogether.

dist/package.sh
```diff
@@ -78,12 +83,23 @@ targets=(
...
 done
 
+
 echo "+ Update version number..."
-sed -i "s/\[\[VERSION\]\]/${VERSION}/g" $DIR/index.html $DIR/troubleshoot/index.html $DIR/dist/app.js $DIR/version.txt
+if [[ "$OSTYPE" == "darwin"* ]]; then
+        sed -i'' -e "s/\[\[VERSION\]\]/${VERSION}/g" $DIR/index.html $DIR/troubleshoot/index.html $DIR/dist/app.js $DIR/version.txt
+    else
+        sed -i "s/\[\[VERSION\]\]/${VERSION}/g" $DIR/index.html $DIR/troubleshoot/index.html $DIR/dist/app.js $DIR/version.txt
```
dbrgn: Same thing here. But careful, you need to add a space between the `i` and the `''`, otherwise it won't work!

docs/self_hosting.md
```diff
@@ -43,6 +43,9 @@ Finally, unpack the `dist/threema-web-[VERSION].tar.gz` archive to your web serv
 Make sure to serve Threema Web only via https. We also recommend to enable
 HSTS, HPKP, CSP and other available security mechanisms in your web server.
 
+#### Notes on building on macOS
+Building on macOS works as well if you have gnu-sed installed. You can get gnu-sed from homebrew with ```brew install gnu-sed```.
```
dbrgn: gnu-sed is not necessary with when not using `gsed`, right?

dist/package.sh
joelfischerr: I was just about to try it and fixed it in https://github.com/threema-ch/threema-web/pull/406/commits/95883b815973c4c91681315260d25d3ddbea1d5a

dist/package.sh
joelfischerr: Fixed in https://github.com/threema-ch/threema-web/pull/406/commits/06bca90125ccc1e6e5c309c07c2c61c85cd46449

dist/package.sh
joelfischerr: Also fixed in https://github.com/threema-ch/threema-web/commit/95883b815973c4c91681315260d25d3ddbea1d5a

docs/self_hosting.md
joelfischerr: Yes, I reverted it in https://github.com/threema-ch/threema-web/pull/406/commits/2c6d3e0a69c52039aa8be2885ae350882be44744

dbrgn:
Great, thanks for the changes @joelfischerr 🙂

Last remaining thing would be the CLA: https://cla-assistant.io/threema-ch/threema-web?pullRequest=406 (Note that it's not a copyright assignment and we may not release your code under a non-open-source license, see 2.3 Outbound License)

Issue #407, opened 2018-01-29
Make it a Progressive Web Application (PWA)


rugk:
It would be nice if you could make Threema Web a Progressive Web Application, so it can be used on a phone as a "app", whcih feels quite native.

Especially as Mozilla recently released Firefox 58 for Android, which now supports This is how PWAs look like:
https://www.youtube.com/watch?v=7sMtsOeqv2w

More information e.g. here: https://developer.mozilla.org/en-US/Apps/Progressive

As Threema Web already operates only offline (static files), the only thing, which may be still missing is a manifest file, so it can be [added to the Homescreen](https://developer.mozilla.org/en-US/Apps/Progressive/Add_to_home_screen).
Okay, you maybe also need some ServiceWorkers to make it possible to launch the PWA without internet connection.

dbrgn:
> Okay, you maybe also need some ServiceWorkers to make it possible to launch the PWA without internet connection.

How would you access the SaltyRTC signaling server without internet connection?

rugk:
How would you use Twitter without a internet connection?
(They also offer a PWA.)

I think the purpose of having the assets locally for a PWA is just to have everything installed and to prevent that it is downloaded again and again (just like a native app) and of course it also starts faster.
Of course PWAs, which can be used completely offline, are optimal, but it maybe is also designed so that it works in bad network conditions, i.e. you only need some small api calls (Twitter) or some tiny connection and otherwise can use LAN connections (Threema Web).

Generally, don't ask me. Ask the ones who invented PWAs. I think the general idea us to make it as close to a native app as possible. And there, the usual concept (which nobody would question) is that you download the app with all assets first, and then only make some connections if necessary.

ovalseven8:
While possible, it's probably not that useful to run Threema Web on a phone. The UX will be bad and the batter drain enormous.

dbrgn:
> Generally, don't ask me.

Well, you requested the feature :) Do you have a compelling use case for this? The only place where this *might* be useful is on a tablet. But as @ovalseven8 said, it could lead to battery drain with the current state of the application.

Anyways, I think PWAs are interesting, but this doesn't have very high priority for us at the moment. Pull requests are welcome though!

rugk:
Okay, added a PR for the manifest. Still, for a real PWA the "offline functionality" is still missing.

As explained in https://infrequently.org/2016/09/what-exactly-makes-something-a-progressive-web-app/: 
[...]

PR #408, opened 2018-01-30
Add WebManifest for "Add to homescreen" feature

rugk:
Ref https://github.com/threema-ch/threema-web/issues/407 (So you can already add it to the Homescreen; at least in Firefox; for Chrome it would need to register some service workers so it can work offline.)

TODO:
* [x] Still need to test this.
* [x] According to [Googles WebManifest guide](https://developers.google.com/web/fundamentals/app-install-banners/) the icon should also be available as 192x192 **png** icon


rugk:
For better favicons in all sizes you may use https://realfavicongenerator.net/. It also, partially, generates the manifest file by itself. (or something like that)

rugk:
So I've tested it with Firefox on Android and it does work there. You may choose a better icon or so, but basically it works.

dbrgn:
Thanks!

What steps did you take to test this in Firefox? I can only see a strange android icon, instead of the "add to home screen" icon that should appear (Firefox 58.0.1).

image

rugk:
And I cannot see that Android icon? Is not this only shown when a website can be opened in another app? That would mean you have an app, which registers localhost...

AFAIK localhost is accepted as a secure origin, but at least I used a webserver, real domain name and HTTPS.

dbrgn:
> AFAIK localhost is accepted as a secure origin, but at least I used a webserver, real domain name and HTTPS.

That's probably the issue :)

dbrgn:
Thanks!

PR #410, opened 2018-01-31
Mention selector in groups, add "@" to mention box

sillych:


rugk:
Great, so other [discussion from here](https://github.com/threema-ch/threema-web/pull/404#issuecomment-360870500) was heard.

Just a question: Why is the @ so grey and why is it a bit like a subscript? (The grey may not be bad, but it  does not really look as if it is on the same line as the username.)

ovalseven8:
Does the selection also work using the keyboard only? :)

sillych:
@ovalseven8 not yet 😺

sillych:
@rugk The mention design is in line with the android app. We fixed the alignment

2018-02-01-065151_443x97_scrot

ovalseven8:
> @ovalseven8 not yet 😺

Oh, please add that. It's a must-have! 😄

sillych:
@ovalseven8 feel free 🥇

ovalseven8:
> @ovalseven8 feel free 🥇

Sorry, I'd like to help here but I lack JavaScript/AngularJS skills. :(

sillych:
Your wish is our command
Keyboard selection implemented:

- Select mention with the up/down arrow keys.
- Press enter to insert to selected mention

ovalseven8:
❤️ @sillych ❤️ 

Thanks! 😉

src/directives/threema_action.ts
```diff
@@ -15,22 +15,38 @@
...
 import {UriService} from '../services/uri';
+import {WebClientService} from '../services/webclient';
 
 export default [
     '$timeout',
     '$state',
     'UriService',
-    function($timeout, $state: ng.ui.IStateService, uriService: UriService) {
+    'WebClientService',
+    function($timeout, $state: ng.ui.IStateService, uriService: UriService, webClientService: WebClientService) {
```
dbrgn: Maybe this directive should be renamed to `click_action` now that it does not just handle threema actions?

src/directives/threema_action.ts
```diff
@@ -76,19 +92,38 @@ export default [
...
+                                        }
+                                    }
+                                    break;
+                                case 'span':
+                                    // supprt only id mentions (not all or me)
+                                    const mentionCssClass = 'mention id';
+                                    const cls = node.getAttribute('class');
+                                    if (cls.substr(0, mentionCssClass.length) === mentionCssClass) {
+                                        // Hack to extract the identity from class name
+                                        const identity = cls.substring(mentionCssClass.length).trim();
```
dbrgn: Instead of using the css class hack, you could use a `data-mention-id` attribute (see https://developer.mozilla.org/en-US/docs/Learn/HTML/Howto/Use_data_attributes). Would make the code much cleaner :)

src/filters.ts
```diff
@@ -139,21 +139,28 @@ angular.module('3ema.filters', [])
...
                             }
                         }
 
-                        if (html !== undefined) {
+                        if (mentionName !== undefined) {
                             text = text.replace(
                                 new RegExp(escapeRegExp(possibleMention), 'g'),
-                                html,
+                                '<span class="mention ' + cssClass + '"'
+                                    + ' text="@[' + identity + ']">' + mentionName + '</span>',
```
dbrgn: We need to be careful about script injection here. Should be fine in this case though, since `identity` can only be a possible mention, so it must match the RegEx.

In the mentionName line breaks (`<br>` tags) are processed, but not things like `<script>`. Also, this is a value you set yourself on your phone, so not much danger :) We should still fix that some time (but not in this PR).

src/partials/messenger.ts
```diff
@@ -557,9 +588,86 @@ class ConversationController {
...
+                this.selectedMention = this.currentMentions.findIndex((m) => {
+                    return m.identity === selectedMentionObject.identity;
+                });
+            }
+        } else {
+            this.currentMentionFilterWord = null;
+        }
+    }
+
+    public getSelectedMention = (): {identity: string, query: string, isAll: boolean} => {
```
dbrgn: Since you use this return type multiple types, you could also define a [type alias](https://stackoverflow.com/questions/21839692/does-typescript-allow-type-aliases#28343437) 🙂

src/partials/messenger.ts
```diff
@@ -557,9 +588,86 @@ class ConversationController {
...
+        }
+
+        return this.currentMentions[this.selectedMention];
+    }
+    /**
+     * Handle mention selector navigation
+     */
+    public onComposeKeyDown = (ev: KeyboardEvent): boolean => {
+        if (this.showMentionSelector() && !ev.shiftKey) {
+            let move = ev.which === 40 ? 1 : (ev.which === 38 ? - 1 : 0);
```
dbrgn: The `KeyboardEvent.which` property is [deprecated](https://developer.mozilla.org/en-US/docs/Web/API/KeyboardEvent/which) and should probably be replaced with [KeyboardEvent.key](https://developer.mozilla.org/en-US/docs/Web/API/KeyboardEvent/key). Same thing on line 650.

src/partials/messenger.ts
```diff
@@ -557,9 +588,86 @@ class ConversationController {
...
+                    this.onMentioned(selectedMentionObject.identity);
+                }
+                return false;
+            }
+        }
+
+        return true;
+    }
+
+    public onMentioned(identity: string = null): void {
```
dbrgn: Maybe we should rename this to `onMentionSelected` to prevent confusion with mentions of yourself? 🙂

dbrgn:
Btw, we should probably disable the "create mention" feature (making it display only) until the new app versions are released.

src/directives/threema_action.ts
sillych: +1

src/directives/threema_action.ts
sillych: The problem is the `ng-bind-html` of angularJS... the directive removes all arguments from the DOMElement (except `class`)

dbrgn: I see :(

dbrgn: I see :( Then maybe add that information to the comment?

rugk:
Can't you generally host a beta version, which one can only use with the beta app on https://web-beta.threema.ch/ and keep the live version on the "stable state"?

src/directives/threema_action.ts
sillych: +1

src/filters.ts
sillych: +1

src/partials/messenger.ts
sillych: 🎉

src/directives/threema_action.ts
sillych: done

sillych:
@dbrgn escaping names in #417

PR #419, opened 2018-02-06
Make mentions readonly for now

dbrgn:
There are still some edge cases with the compose area on Firefox.


rugk:
Has this been reverted by now? Mentions still do not work in v1.8.2-gh.

PR #420, opened 2018-02-07
add pseudo element arround a mention to fix the remove bug in ff

sillych:


dbrgn:
Thanks! Two bugs I found when testing:

- When the mention is at the beginning of the input field, I cannot move before it with the arrow keys. Similarly, when it's at the end of the input field and the cursor is before the mention, I cannot move to the end of the input field with the arrow keys.
- If the browser window height is very small, and if there are a lot of members in the group, the compose area disappears when entering `@`. There should be a scrollbar instead. (I can send you a screenshot)

sillych:
> If the browser window height is very small, and if there are a lot of members in the group, the compose area disappears when entering @. There should be a scrollbar instead. (I can send you a screenshot)
Good catch, the same behavior with quotes 👎

src/filters.ts
```diff
@@ -149,7 +149,7 @@ angular.module('3ema.filters', [])
                             cssClass = 'me';
                         } else {
                             const contact = webClientService.contacts.get(possibleMention.substr(2, 8));
-                            if (contact !== null) {
+                            if (contact !== null && contact !== undefined) {
```
dbrgn: Oh JS :sob:

rugk: don't know about typecrypt, but when comparing to `undefined` kit is, AFAIK, usually recommend to compare the [type to the "one of" undefined](https://stackoverflow.com/questions/2778901/how-to-compare-variables-to-undefined-if-i-don-t-know-whether-they-exist#2779104), i.e., add `typeof` and compare it as a string.

dbrgn: @rugk That is obsolete with ES2015.

rugk: Great to hear that! 😄

lgrahl:
@dbrgn Is this still an issue we need to look at?

dbrgn:
Yes. @sillych do you want to give this another try?

dbrgn:
Closing in favor of the compose area rewrite.

PR #425, opened 2018-02-11
Prevent drag of background image

rugk:
There are many ways to do this (https://stackoverflow.com/questions/4211909/disable-dragging-an-image-from-an-html-page), so I just chose this one. It may also prevent other useless things of the background image.

Tested in Firefox 58.

### DIFF
--- a/src/sass/layout/_main.scss
+++ b/src/sass/layout/_main.scss
@@ -10,6 +10,7 @@
     -moz-user-select: none;
     -webkit-user-select: none;
     user-select: none;
+    pointer-events: none;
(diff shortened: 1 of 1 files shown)


ovalseven8:
For me it's only possible to "drag" the background image, but I couldn't "drop" and send it. (FF 58)

rugk:
Okay, so changed the title... 😆 
I actually have not tested the drag, it's just that one usually uses both terms together. 😃

dbrgn:
Nice, thanks!

PR #436, opened 2018-02-19
Add gray background to download icon

ovalseven8:
Old:
bildschirmfoto_2018-02-19_10-36-34
New:
bildschirmfoto_2018-02-19_09-44-01

I am not sure, if I put the changes into the right place. In my example it affected the `i` element, not ` img`. However, it was defined together so I  put it there. Please adjust if needed!

### DIFF
--- a/src/sass/components/_message_media.scss
+++ b/src/sass/components/_message_media.scss
@@ -43,6 +43,8 @@ $loading-ring-thickness: 5px;
             color: white;
             margin: auto auto;
             position: relative;
+            background: rgba(128, 128, 128, .3);
+            border-radius: $material-radius;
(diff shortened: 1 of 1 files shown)


src/sass/components/_message_media.scss
```diff
@@ -43,6 +43,8 @@ $loading-ring-thickness: 5px;
             color: white;
             margin: auto auto;
             position: relative;
+            background: rgba(128, 128, 128, .3);
+            border-radius: 5px;
```
sillych: please use `$material-radius` instead `5px`

ovalseven8:
Okay, done.

PR #437, opened 2018-02-19
Add gray background to video play icon

ovalseven8:
Old:
b
New:
a

It makes the play icon visible if the video thumbnail is white.
Not sure if I put the changes to the right place, please check.

### DIFF
--- a/src/sass/components/_message_media.scss
+++ b/src/sass/components/_message_media.scss
@@ -122,6 +122,8 @@ $loading-ring-thickness: 5px;
                     position: absolute;
                     top: calc(50% - #{$circle-size/2});
                     left: calc(50% - #{$circle-size/2});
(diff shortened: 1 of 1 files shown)


src/sass/components/_message_media.scss
```diff
@@ -122,6 +122,8 @@ $loading-ring-thickness: 5px;
                     position: absolute;
                     top: calc(50% - #{$circle-size/2});
                     left: calc(50% - #{$circle-size/2});
+                    background: rgba(128, 128, 128, 0.6);
+                    border-radius: 40px;
```
sillych: try `$circle-size - 2 * $loading-ring-thickness` instead `40px`

ovalseven8: Done.

PR #438, opened 2018-02-19
Change cursor to pointer

### LINKED ISSUE #435: Change cursor to 'pointer' to indicate that you can click on group
## Expected Behavior
Cursor should be of type ` pointer` to show the user that you can click on the group.

## Current Behavior
bildschirmfoto_2018-02-19_09-11-36

1. Go to a contact page of someone you're in a group with
2. At the bottom you can find all the groups you both are in
3. You can click on those, however it's not clear that you can because of missing 'pointer'

ovalseven8:
Is this the right class to put it in?

Fixes #435

### DIFF
--- a/src/partials/messenger.receiver/contact.html
+++ b/src/partials/messenger.receiver/contact.html
@@ -85,7 +85,7 @@
 		<md-card-content>
 			<ul class="group-list">
 				<li ng-repeat="groupReceiver in ctrl.inGroups">
-					<eee-group-badge eee-group-receiver="groupReceiver" eee-contact-receiver="ctrl.receiver"/>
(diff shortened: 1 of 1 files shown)


dbrgn:
Thanks!

> Is this the right class to put it in?

Not quite, since there can be receiver badges that are not clickable (for example the own badge in the group details view).

Instead, you can simply add `class="linked"` to the `eee-group-badge` in `src/partials/messenger.receiver/contact.html`, then the mouse hand should appear (thanks to [this](https://github.com/ovalseven8/threema-web/blob/475a954228652781d670af171931cfea598ed8ae/src/sass/components/_receiver_badge.scss#L71-L73) rule).

ovalseven8:
Better now? 😕

PR #439, opened 2018-02-20
Fix supportsPassive helper function

dbrgn:
Due to missing type annotations, we did not notice that no value was
returned. Thus passive event listeners were always turned off.

Fixing the detection might improve scroll performance on newer browsers.
See https://github.com/WICG/EventListenerOptions/blob/gh-pages/explainer.md


dbrgn:
Tests should pass once #440 is merged.

PR #443, opened 2018-02-22
Fix duplicate in changelog

ovalseven8:


dbrgn:
Thanks :D I also just noticed that I forgot to add you as a contributor. Will fix that :slightly_smiling_face:

PR #448, opened 2018-02-22
Fix some spellings

ovalseven8:
- Change "Github" to "GitHub"
- Change "Already Connected" to "Already connected" (adjective uncapitalized)
- Change "..." to "…" with a leading space (own character; the "main" change is the leading space that is needed)

Sorry for the "wisenheimer"-changes. Just ignore it and close the PR if you like it as it is. :smile: 

I just noticed this every time I saw it. ;)

### DIFF
--- a/public/i18n/de.json
+++ b/public/i18n/de.json
@@ -24,9 +24,9 @@
         "BROWSER_NOT_SUPPORTED": "Dieser Browser wird nicht unterst\u00fctzt",
         "ANDROID_ONLY": "Kompatibel mit Threema für Android.",
         "CONNECTING": "Verbindung wird aufgebaut",
-        "WAITING_FOR_PUSH": "Threema-App wird<br>aufgeweckt...",
-        "CONNECTING_TO_APP": "Verbindung zu App<br>wird aufgebaut...",
-        "LOADING_INITIAL_DATA": "Daten werden geladen...",
+        "WAITING_FOR_PUSH": "Threema-App wird<br>aufgeweckt …",
+        "CONNECTING_TO_APP": "Verbindung zu App<br>wird aufgebaut …",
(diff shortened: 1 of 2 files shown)


public/i18n/de.json
```diff
@@ -24,9 +24,9 @@
         "BROWSER_NOT_SUPPORTED": "Dieser Browser wird nicht unterst\u00fctzt",
         "ANDROID_ONLY": "Kompatibel mit Threema für Android.",
         "CONNECTING": "Verbindung wird aufgebaut",
-        "WAITING_FOR_PUSH": "Threema-App wird<br>aufgeweckt...",
-        "CONNECTING_TO_APP": "Verbindung zu App<br>wird aufgebaut...",
-        "LOADING_INITIAL_DATA": "Daten werden geladen...",
+        "WAITING_FOR_PUSH": "Threema-App wird<br>aufgeweckt …",
```
rugk: I would not use a leading space here. IMHO that is not common…

(example above)

That way it looks strange …. You usually never use a space , before such a character, because the last comma here looks much better.

ovalseven8: Well, it is common actually. When a sentence is ending, you usually put a space before the three dots (http://www.neue-rechtschreibung.net/2009/11/17/vom-richtigen-umgang-mit-drei-punktchen/). That's the rule at least.

However, yes, you'll find enough examples without a space. :smile: 

Let's see what the Threema guys say about this. ;)

rugk: Okay, never knew this. https://de.wikipedia.org/wiki/Auslassungspunkte actually provs your point. Kinda uncommon, but well …

dbrgn: TIL :slightly_smiling_face: Wasn't aware of that rule either.

PR #456, opened 2018-03-02
Return a WordResult object

### LINKED ISSUE #454: Bug while converting emoji shortnames
## Expected Behavior
`:smile:` -> :smile: 

## Current Behavior
If you use the blank key after you wrote `:smile:` very quickly, you get ::smile: as result (notice the leading colon). If you wait a few deciseconds after you wrote `:smile:` everything is okay.

Can you reproduce?

* Threema Web version: 1.8.2
* Browser name and version: Firefox 58.0.2

sillych:
fix #454

### DIFF
--- a/src/threema.d.ts
+++ b/src/threema.d.ts
@@ -496,7 +496,9 @@ declare namespace threema {
     }
 
     interface WordResult {
+        // The trimmed word
         word: string;
+        // The length of the untrimmed word
         realLength: number;
     }
 
--- a/src/threema.d.ts
+++ b/src/threema.d.ts
@@ -148,7 +148,7 @@ declare namespace threema {
 
     const enum IdentityType {
         Regular = 0,
-        Work
+        Work,
     }
 
     /**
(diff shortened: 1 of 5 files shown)


dbrgn:
> fox #454

:fox_face:

sillych:
:sweat_smile:

src/threema.d.ts
```diff
@@ -495,6 +495,11 @@ declare namespace threema {
         isAll: boolean;
     }
 
+    interface WordResult {
+        word: string;
+        realLength: number;
```
dbrgn: What exactly does `realLength` mean? From the tests it looks like it also includes trailing whitespace?

sillych: Length of the untrimmed word

dbrgn: I pushed a commit with some code comments: https://github.com/threema-ch/threema-web/pull/456/commits/22dde4b6c396acefd5c4f662aef2f8461184a321 Are they correct?

sillych: :1st_place_medal:

Issue #198, opened 2017-04-18
Sync drafts


rugk:
## Expected Behavior

When writing drafts on the web or in the app these should be synced so one can continue to write them on the other device…

## Current Behavior

The app and the web app maintain their own drafts and one can only continue to write the draft on the "right" device.

## Possible Solution

Sync, …

PR #475, opened 2018-03-15
Protocol v2 / iOS Support [Beta]

### LINKED ISSUE #4: Show/Hide Inactive Contacts
Add an option to enable/disable display of inactive contacts.

No app/protocol update is necessary, the flag is already sent with a contact.

### LINKED ISSUE #32: Allways display sent/received time
since the display has more space, I would love to see both times in the chat log

### LINKED ISSUE #58: iOS Support Request
iOS Support Request Please.

### LINKED ISSUE #97: Single emojis should be large
## Expected Behavior

When sending only one single emoji, the emoji should be displayed largely to be consistent with the App.

## Current Behavior

The emoji is displayed in the normal font size

## Possible Solution

Check if a message only contains a single emoji (watch out for [emoji modifiers](http://unicode.org/reports/tr51/#Emoji_Modifiers_Table) like skin tones and [ZWJ](http://unicode.org/reports/tr51/#Emoji_ZWJ_Sequences)s which could count as multiple symbols, but represent a single emoji) and set the font size to a higher amount.

1. Send a single emoji as a message
2. Compare the displayed message on Threema Web and the Android app

## Your Environment

* Threema Web version: 1.0.3
* Threema for Android version: 3.02
* Browser name and version: Chrome 56
* Operating system and version: macOS 10.12

### LINKED ISSUE #154: Saving profile without setting picture resets picture
## Expected Behavior

When changing the name of a person via Web, the profile picture should not be changed.

## Current Behavior

When saving a profile without setting a profile picture, it's removed.

1. Edit a profile
2. Add a profile picture
3. Save, everything works
4. Edit the profile again
5. Save, profile picture is gone

### LINKED ISSUE #221: Changing own avatar
It would be nice if one could also change/set ones profile picture in the web UI too, so one can use images on your computer('s hard drive) – similar to changing the picture of other contacts, which is already possible.

BTW: Maybe also display the profile picture somewhere in the UI…

### LINKED ISSUE #393: Broken conversation preview when deleting only message in conversation

1. Start a new conversation
2. Send a message
3. Delete that message again
4. Conversation preview now looks like this:

img

### LINKED ISSUE #445: Race condition in password field
## Expected Behavior

When reconnecting a session, the password unlocks the session data. The same password for unlocking should also be used to store session data.

## Current Behavior

There seems to be a race condition when reading the password field's value. The password that unlocks session data can be different from the password that stores session data because the field is still writable after having pressed *Reconnecting* (or the Enter key).

## Possible Solution

Read the password field's value once and disable it immediately.

1. Unlock a session by typing your password and press Enter.
2. Immediately press another key.
3. Press F5. Your session will be locked by the *new* password.

### LINKED ISSUE #476: Add: Contact is writing status
The "status" "… is writing." should be added, so I can see when someone writes a message.

It should also be shown in the contacts overview, as it is done in the Threema app.

### LINKED ISSUE #480: Laggy input
## Expected Behavior
No input latency.

## Current Behavior
Laggy input.

1. Go to https://pixabay.com/de/s%C3%A4ugetiere-wildlife-teuer-fox-3218028/
2. Right click -> Copy image
3. Go to a Threema Web conversation (e.g. ECHOECHO)
4. Paste image using Ctrl+V
5. Write something in the image description (type fast, so the problem is more visible)
6. Extremely laggy input

Now try the following:

1. Go to https://de.wikipedia.org/wiki/Threema#/media/File:Threema.png
2. Right click -> Copy image
3. Go to a Threema Web conversation (e.g. ECHOECHO)
4. Paste image using Ctrl+V
5. Write something and you should notice that the input is faster now. So it seems like it has something to do with the media (perhaps media size?).

## Your Environment
* Threema Web version: 1.8.2
* Threema for Android version: 3.41
* Browser name and version: Firefox 59.0.2

### LINKED ISSUE #507: Still get browser notifications for "do not disturb" chats
## Expected Behavior

No browser notifications. (as it also is done correctly in the app, where I get no notification)

## Current Behavior

I still get browser notifications, just without a sound. (as if they were just "muted")

1. Mark a (group) chat in Threema as "do not disturb (unless you are @mentioned). And I don't mean the "mute" setting by it – this can stay the default.
2. Get a message in that chat while connected to Threema Web.

## Your Environment

* Threema Web version: 1.8.2-gh 
* Threema App version: 3.41
* Browser name and version: Firefox 60
* Computer operating system and version: Fedora 28 (GNOME)
* Smartphone operating system and version: Android

### LINKED ISSUE #517: Feedback when "copy to clipboard"
## Expected Behavior

As the Threma app does, show a notification somewhere "text copied" or so.

## Current Behavior

The menu just vanishes. I doubt it is enough to signal to the user that the process worked.

1. Copy a text.

### LINKED ISSUE #519: Flashing battery icon is CPU intensive
Turns out the pulsating battery icon is very CPU intensive, at least in Firefox. Weirdly...

dbrgn:
Tracking issue, work in progress.

Will fix #58 / fix #221 / fix #154 / fix #476 / fix #393 / fix #32 / fix #97 / fix #445 / fix #507 / fix #4 / fix #517 / fix #519 / fix #480.


ovalseven8:
May I ask what are the protocol changes? :-)

ovalseven8:
Well, perhaps I've asked the wrong question. :smile: 

Of course, I can find some information from the PR. What I wanted to know is what those protocol changes (beside iOS support) will change/improve. Or are those changes just small things like changing the date format without further effect on stability/performance etc.?

Sorry.

rugk:
Ah, so you wanted to ask about the PR and not ask for things by yourself. Sorry, misunderstood your question.

dbrgn:
> May I ask what are the protocol changes? :-)

Nothing fancy, just a lot of cleanup to a historically grown protocol (e.g. moving success flags from data to args, sending dates as timestamp, etc).

public/i18n/de.json
```diff
@@ -176,8 +176,22 @@
...
+            "badRequest": "Ungültige Anfrage (Protokollfehler?)",
+            "timeout": "Timeout",
+            "internalError": "Ein interner Fehler ist aufgetreten",
+            "invalidIdentity": "Ungültige Threema-ID",
+            "invalidContact": "Ungültiger Kontakt",
+            "invalidGroup": "Ungültige Gruppe",
+            "invalidDistributionList": "Ungültige Verteilerliste",
+            "notAllowed": "Bearbeitung nicht erlaubt",
+            "notAllowedLinked": "Systemkontakt kann nicht verändert werden",
+            "notAllowedBusiness": "Avatar eines Threema Work Kontaktes kann nicht geändert werden",
```
sillych: A business contact is a "gateway contact" not a "work contact"

dbrgn: Ah, good catch :)

N3dal:
How about adding support for #198 in Protocol v2 ?

joelfischerr:
Why has the compatibility footer been removed?
It might be confusing for people who have heard that Threema Web is only available for Android but have missed that it's also available for iOS.
And the Windows Phone version will still be available after the release of Threema Web for iOS but won't have Threema Web support, which might add some more confusion.

dbrgn:
@joelfischerr It's a bit more complicated than that, since for example Safari will only support iOS, but not Android in the beginning. So we'll show browser-specific hints instead. Also, Windows Phone is dying at a slow but steady pace, so most people using Threema on that platform will probably be aware that not all features are implemented :)

Ruko2010:
Hi, is there an ETA for iOS Support for Threema Web?

joelfischerr:
Yes, there is a post on their blog: https://threema.ch/en/blog/posts/threemas-web-client-ios

> We are close to the finish line, but there is still quite some work ahead. We’re anticipating a public beta by May.

But for general questions about Threema or Threema Web it might be better to use threema-forum.de.

dbrgn:
@Ruko2010 since the announcement says "by May": Within a week :slightly_smiling_face:

Ruko2010:
@joelfischerr sorry, I wasn't aware that there is a forum.
@dbrgn I did not notice this little sentence 😃. I'm thrilled to use the beta.

Thanks for your work!

dbrgn:
@ovalseven8 on iOS you only get 30 seconds of connectivity per push. Threema Web will automatically reconnect when an event happens on one of the two sides. So connection loss means "waiting to reconnect as soon as something happens".

ovalseven8:
Interesting because I thought Threema calls had been introduced to be able to make persistent connections also on iOS. But as long as this works fine, it's actually a better solution because this should reduce battery comsumption, right?

dbrgn:
Not the persistent connection, but the background pushes. They are not possible at all without voip push tokens.

dbrgn:
Note: This is now merged into master (since Android now also uses the protocol version 2), but iOS will stay in beta right now.

PR #536, opened 2018-07-30
Avoid hiding disconnect reason dialog on stop

### LINKED ISSUE #529: Disconnect reason msg disappears
- Open a new non-persistent connection
- Delete it in Android
- Threema Web quickly shows a message box and hides it again

(Beta)

dbrgn:
The stop() call above may result in a redirect, which will in turn hide
all open dialog boxes. Therefore, to avoid immediately hiding the alert
box, enqueue dialog at end of event loop.

Fixes #529.

### DIFF
--- a/src/services/webclient.ts
+++ b/src/services/webclient.ts
@@ -1717,7 +1717,7 @@ export class WebClientService {
         }
         const reason = message.data.reason;
 
-        this.$log.debug(this.logTag, 'Disconnecting requested (reason: ' + reason + ')');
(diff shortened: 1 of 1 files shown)


src/services/webclient.ts
```diff
@@ -1717,7 +1717,7 @@ export class WebClientService {
         }
         const reason = message.data.reason;
 
-        this.$log.debug(this.logTag, 'Disconnecting requested: reason=', reason);
```
lgrahl: You can use [template literals](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Template_literals) here. :)

PR #558, opened 2018-08-13
Fix work indicator

dbrgn:
Due to a bug in the Android app, the `isWork` flag was not always defined.

### DIFF
--- a/src/services/webclient.ts
+++ b/src/services/webclient.ts
@@ -2410,7 +2410,7 @@ export class WebClientService {
             device: data.device,
             os: data.os,
             osVersion: data.osVersion,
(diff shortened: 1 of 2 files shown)


src/services/webclient.ts
```diff
@@ -2410,7 +2410,7 @@ export class WebClientService {
             device: data.device,
             os: data.os,
             osVersion: data.osVersion,
-            isWork: data.isWork,
+            isWork: hasValue(data.isWork) ? data.isWork : false,  // Backwards compat hack, remove after August 2019
```
dbrgn: This is fixed in master, but let's leave the fix in here for a while.

lgrahl: Nit: I would add this as a `TODO`. Easier to track.

PR #559, opened 2018-08-13
#546 - Makes buttons more accessible

SirTyson:
Fixes #546. Buttons for attach file, emoji, and send are now labeled as buttons for accessibility readers.

Issue #172, opened 2017-03-14
Keyboard Shortcuts


dbrgn:
## Expected Behavior

The user should be able to use keyboard shortcuts for often used things.

## Possible Solution

- Search: `/`
- Conversations tab: `1`
- Contacts tab: `2`
- Cheatsheet: `?`
- Close search: `ESC`
- Close session: `q`

https://chieffancypants.github.io/angular-hotkeys/

rugk:
For search please support `Ctrl+F`, which is very common.

dbrgn:
> For search please support Ctrl+F, which is very common.

And it's already bound in the browser itself. We don't want to override built-in shortcuts, that's extremely annoying. (Discourse does it for example, and I hate it.)

ovalseven8:
Duplicate of #37?

fichtennadel:
Please also add shortcut/preview for emoticons like for example in in slash by typing `:<letter>` shows preview of available emoticons:

![image](https://cloud.githubusercontent.com/assets/2324670/24006280/9f636232-0a6b-11e7-8e0a-8a9228c711e6.png)

dbrgn:
@fichtennadel That's another issue :) #1

rugk:
>  (Discourse does it for example, and I hate it.)

THat's very subjective. Actually I often like it very much. And Ctrl+F could be useful - or maybe Ctrl+Shift+F if you want… Or make it configurable…

dbrgn:
For accessibility reasons, it should be possible to navigate between left navigation, conversation header, conversation and compose area using a keyboard shortcut.

In #562, @MarcoZehe suggested Ctrl+F6. That does seem like a de-facto standard for switching between different areas it seems: https://defkey.com/what-means/ctrl-f6 Slack also uses that shortcut.

PR #562, opened 2018-08-14
Improve Accessibility

### LINKED ISSUE #546: Accessibility: Make Send and Attach File buttons real buttons
## Expected Behavior

When using Threema Web with a screen reader, both items Send and Attach Files should be buttons. 

## Current Behavior

Currently, screen readers only announce them as clickable.

## Possible Solution

Either make them real `<button>` elements, or add `role="button" tabindex="0"` and a keyPress handler that does the same as the click handler and reacts to Space and Enter.

1. Connect to ThreemaWeb with Firefox + NVDA (on Windows), or Safari or Chrome with VoiceOver turned on (on Mac).
2. Open a chat.
3. Find the Send and Attach Files button near the end of the output, after the input box.
4. Listen to what the screen reader says. In Firefox, you can also examine the Accessibility Tree with the new Accessibility Inspector, enabled in Dev Tools Settings.

## Your Environment
* Threema Web version: 2.0.0.8 Beta
* Threema App version: 4.0 Beta (iOS)
* Browser name and version: Firefox 61.0.1 on Windows
* Computer operating system and version: Windows 10 1803 (April 2018 update)
* Smartphone operating system and version: iOS 12 Public Beta 5, but also happens with 11.4.1.

dbrgn:
Add a lot of ARIA attributes for improved accessibility

Tested with macOS voiceover in Safari.

There's still a lot more to do, but it's a start. For example, the navigation on the left and the conversation should be hierarchical groups, otherwise it's not possible to escape the conversation, because by going back through the elements, new messages are constantly loaded and inserted into the voiceover navigation stack. I'm not yet sure how to do that, documentation / reference material on ARIA seems really bad.

Fixes #546, replaces #559.

@MarcoZehe if you have additional feedback on these changes, that would be welcome!

### DIFF
--- a/index.html
+++ b/index.html
@@ -57,7 +57,7 @@
 </head>
 
 <body ng-controller="StatusController as ctrl" class="{{ ctrl.statusClass }}" ng-class="{expanded: ctrl.expandStatusBar}">
-    <img src="img/bg.jpg?v=1" id="background-image" draggable="false">
+    <img src="img/bg.jpg?v=1" alt="Background image: Blurred photo of a mountain" id="background-image" draggable="false">
 
     <noscript>
         <img id="logo-noscript" src="img/logo.svg?v=[[VERSION]]"/>
@@ -69,7 +69,7 @@
 
     <div id="main-wrapper" ng-cloak ng-class="{wide: ctrl.wide()}">
         <header>
-            <h1 id="title">
+            <h1 id="title" aria-label="Threema Web Logo">
                 <div id="logo" ng-include src="'img/logo.svg?v=[[VERSION]]'"></div>
             </h1>
         </header>
--- a/index.html
+++ b/index.html
@@ -91,11 +91,8 @@
         </div>
         <footer ng-controller="FooterController as ctrl">
             <ul>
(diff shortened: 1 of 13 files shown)


dbrgn:
> The only way to solve this is to provide a keyboard shortcut that will set focus to the proper places.

That will be application specific though. I doubt users are better off if they have to remember shortcuts for every application. For a use case like that, wouldn't a browser extension like vimium (with it's "f" key) be the better solution?

rugk:
Don't understand the focus problem. Just add `tabindex`. You do focus with tab, of course. And that is useful to anyone, who is faster with the keyboard than mouse. (Try it on GitHub right here, it works quite well...)
(That you may need to send with space or enter is a different problem, but that's a different problem.)

rugk:
A nice thing GitHub has is a "hidden skip to content" button. You can only show it by focusing it and it allows you to skip all that header stuff:

grafik

dbrgn:
It's definitely far from perfect, but this PR improves a lot of things. I'll merge it for now, the area focus issue can be handled in #172.

PR #567, opened 2018-08-20
WIP: Improve UI Performance

dbrgn:
This is still work in progress.

I added a filter profiler that shows the call count & total duration for AngularJS filters. It's not pretty but it works for improving performance.

Initially (simply loading all my conversations):

filters1

After processing text early in MessageText directive:

filters2

By not calculating the filters every digest cycle, we get the following call count improvements:

- emojify (548x -> 426x, 7ms -> 5ms)
- linkify (240x -> 120x, 107ms -> 26ms)
- escapeHtml (548 -> 426x, 6ms -> 3ms)

(Filters with less than 5ms total time aren't listed here)

More improvements are planned. In general, template filters should be avoided.

### DIFF
--- a/index.html
+++ b/index.html
@@ -67,6 +67,32 @@
         </div>
     </noscript>
 
+    <div id="filters">
+        emojify: <span class="emojify"></span>
+        | linkify: <span class="linkify"></span>
+        | markify: <span class="markify"></span>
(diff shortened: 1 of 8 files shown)


dbrgn:
Closing this for now. The important changes from here are now in master, we can always continue working on this branch later.

PR #569, opened 2018-08-21
Refactor browser detection

dbrgn:
Since we won't support the buggy Safari 10 data channel implementation with Android for the time being, we can also add Safari 10 support (e.g. on older iPads) for connection with iOS devices.

What this change does is hiding the "browser is not supported" warning in the beginning. There is still a note saying that Safari is not compatible with Android devices.

Additionally, Firefox and Chrome on iOS are explicitly not supported.

### DIFF
--- a/src/app.ts
+++ b/src/app.ts
@@ -73,7 +73,7 @@ angular.module('3ema', [
     FF: 50,
     CHROME: 45,
     OPERA: 32,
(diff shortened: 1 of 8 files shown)


dbrgn:
> I would have moved browser detection logic into the class itself though.

Yeah, could still be done later :slightly_smiling_face:

PR #573, opened 2018-08-21
Fix size of buttons on connect error page

dbrgn:
The buttons on the connect error page were fine in English...

screenshot

...but not in German:

screenshot: buttons too small

I made them a bit larger:

screenshot: fixed

### DIFF
--- a/src/sass/sections/_welcome.scss
+++ b/src/sass/sections/_welcome.scss
@@ -153,11 +153,11 @@
             .forget {
                 margin-top: 8px;
             }
-            
+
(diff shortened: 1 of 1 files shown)


rugk:
There are other languages, too. So better just not hardcode the sizes… That's almost never a good idea. :wink:

dbrgn:
> There are other languages, too

But not in the codebase right now.

> So better just not hardcode the sizes…

The size *is* limited by the container.

dbrgn:
That works too.

PR #586, opened 2018-09-10
Handshake Protocol Fixes

lgrahl:
There were a couple of hard to find bugs...

### DIFF
--- a/src/services/webclient.ts
+++ b/src/services/webclient.ts
@@ -2189,12 +2189,23 @@ export class WebClientService {
         }
 
         // Stop and show an alert on the welcome page
+        const isWelcome = this.$state.includes('welcome');
         this.stop({
             reason: reason,
             send: false,
             // TODO: Use welcome.{reason} once we have it
             close: 'welcome',
         });
+
+        // Note: This is required to reset the mode and potentially
+        //       re-establish a connection if needed.
+        // TODO: Remove once we have created pages for each mode on the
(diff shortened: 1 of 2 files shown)


src/services/webclient.ts
```diff
@@ -1209,20 +1211,21 @@ export class WebClientService {
...
+            const ready = routine.requiredSteps.every((requiredStep) => {
+                return this.initialized.has(requiredStep);
+            });
+
+            if (ready) {
                 this.$log.debug(this.logTag, 'Running routine after initialization "' + name + '" completed');
+
+                // Important: Remove the routine BEFORE calling it to prevent
+                //            it from being called more than once (due to nested
+                //            calls to .registerInitializationStep).
```
dbrgn: Nice catch!

src/services/webclient.ts
```diff
@@ -175,7 +175,7 @@ export class WebClientService {
     // State handling
     private startupPromise: ng.IDeferred<{}> = null; // TODO: deferred type
     private startupDone: boolean = false;
-    private pendingInitializationStepRoutines: threema.InitializationStepRoutine[] = [];
+    private pendingInitializationStepRoutines: Set<threema.InitializationStepRoutine> = new Set();
```
dbrgn: This change removes the property that the order of the callbacks is not retained. But I don't think that matters for us, so that's fine :slightly_smiling_face:

PR #588, opened 2018-09-17
Don't linkify latest message excerpt in conversation list

### LINKED ISSUE #544: Threema Web shows link to media files although it's not really a link to those
## Expected Behavior
Do not show a link to media files here because actually it's just for chat switching. Remove link and display it like in the app.

## Current Behavior

Not really a bug, but a __minor__ UI flaw IMHO. 

bildschirmfoto_2018-08-03_08-08-16

## Your Environment
* Threema Web version: 2.0.0
* Threema App version: 3.50
* Browser name and version: Firefox 61.0.1

dbrgn:
There's no need for linkifying text in the conversation list.

Fixes #544.

### DIFF
--- a/src/directives/message_text.ts
+++ b/src/directives/message_text.ts
@@ -26,7 +26,8 @@ export default [
             scope: {},
             bindToController: {
                 message: '=eeeMessage',
-                multiLine: '=?eeeMultiLine',
+                multiLine: '@?multiLine',
+                linkify: '@?linkify',
             },
             controllerAs: 'ctrl',
             controller: ['WebClientService', '$filter', function(webClientService: WebClientService, $filter: ng.IFilterService) {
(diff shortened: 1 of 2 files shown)


src/directives/message_text.ts
```diff
@@ -59,17 +60,22 @@ export default [
...
                 this.enlargeSingleEmoji = webClientService.appConfig.largeSingleEmoji;
 
                 this.$onInit = function() {
-                    if (this.multiLine === undefined) {
-                        this.multiLine = true;
-                    }
-                    this.text = processText(getText(this.message), this.largeSingleEmoji, this.multiLine);
+                    // Because this.multiLine and this.linkify are bound using an `@` binding,
+                    // they are either undefined or a string. Convert to boolean.
+                    const multiLine = (this.multiLine === undefined || this.multiLine === 'true');
```
sillych: Why not `const` multiLine = (this.multiLine === undefined || this.multiLine !== 'false');`, because `true` should be the default....

PR #601, opened 2018-10-04
Implement chat pinning

### LINKED ISSUE #361: Mark chats feature missing
The new "mark chats" feature is missing on Threema web.
This includes the indicator and the ability to mark / unmark chats.
It is not a "must have" feature, but it would be quite nice.

dbrgn:
Fixes #361.

Review & testing welcome, @sillych @lgrahl!

### DIFF
--- a/public/img/ic_pin.svg
+++ b/public/img/ic_pin.svg
@@ -1,0 +1,10 @@
+<?xml version="1.0" encoding="UTF-8" standalone="no"?>
+<svg xmlns="http://www.w3.org/2000/svg"
+  height="24"
+  width="24"
+  viewBox="0 0 24 24">
+  <path
+    fill="#000000"
+    fill-opacity="0.54"
+    d="M16,12V4H17V2H7V4H8V12L6,14V16H11.2V22H12.8V16H18V14L16,12Z"/>
+</svg>
--- a/public/img/ic_pin.svg
+++ b/public/img/ic_pin.svg
@@ -5,6 +5,6 @@
   viewBox="0 0 24 24">
   <path
     fill="#000000"
-    fill-opacity="0.54"
+    fill-opacity="0.87"
     d="M16,12V4H17V2H7V4H8V12L6,14V16H11.2V22H12.8V16H18V14L16,12Z"/>
 </svg>
--- a/src/helpers.ts
+++ b/src/helpers.ts
@@ -371,3 +371,17 @@ export function isActionTrigger(ev: KeyboardEvent): boolean {
             return false;
     }
 }
+
+/*
+ * Create a shallow copy of an object.
+ */
+export function copyShallow<T extends object>(obj: T): T {
+    return Object.assign({}, obj);
+}
+
+/**
+ * Create a deep copy of an object by serializing and deserializing it.
+ */
+export function copyDeep<T extends object>(obj: T): T {
+    return JSON.parse(JSON.stringify(obj));
+}
--- a/src/partials/messenger.conversation.html
+++ b/src/partials/messenger.conversation.html
@@ -6,6 +6,8 @@
 
     <div id="conversation-header" class="detail-header">
         <ng-include src="'partials/messenger.backbutton.html'"></ng-include>
+
+        <!-- Conversation details -->
         <div class="header-avatar" ng-click="ctrl.showReceiver()">
             <eee-avatar eee-receiver="ctrl.receiver"
                         eee-resolution="'low'"></eee-avatar>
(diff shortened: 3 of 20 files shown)


src/helpers.ts
```diff
@@ -371,3 +371,17 @@ export function isActionTrigger(ev: KeyboardEvent): boolean {
...
+ */
+export function copyShallow<T extends object>(obj: T): T {
+    return Object.assign({}, obj);
+}
+
+/**
+ * Create a deep copy of an object by serializing and deserializing it.
+ */
+export function copyDeep<T extends object>(obj: T): T {
+    return JSON.parse(JSON.stringify(obj));
```
lgrahl: I'm still puzzled by JS not having an exposed structured clone function... I'll add it to my list of JS quirks, just next to *everything is a float, deal with it... oh, and forget about u64*.

src/partials/messenger.conversation.html
```diff
@@ -22,7 +24,28 @@
                 <span ng-bind-html="ctrl.receiver.members | idsToNames | escapeHtml | emojify"></span>
             </div>
         </div>
+
+        <!-- Menu -->
+        <div class="header-buttons">
```
lgrahl: I would prefer this being an unordered list that grows to the left but we can do that when we add more items.

src/partials/messenger.conversation.html
```diff
@@ -22,7 +24,28 @@
...
+            <md-button
```
lgrahl: Mhh... would it make sense to move this into a directive? The two buttons seem awkward to me. Perhaps this could even be reused for further toggle buttons.

src/partials/messenger.ts
```diff
@@ -844,6 +854,26 @@ class ConversationController {
...
     }
+
+    /**
+     * Mark the current conversation as pinned.
+     */
+    public pinConversation(): void {
+        this.webClientService
+            .modifyConversation(this.conversation, true)
+            .then(() => this.showMessage('messenger.PINNED_CONVERSATION_OK'))
+            .catch((msg) => this.$log.error(this.logTag, 'Pinning conversation failed: ' + msg));
```
lgrahl: I'm wondering if we should also show toast messages for these kinds of responses.

Pro: It clearly is a response to the user's action and let's it know that something it requested didn't work.
Con: It is a bug if it doesn't work and showing the toast as well as adding translations for it is a bit of overhead.

src/services/webclient.ts
```diff
@@ -1812,6 +1813,33 @@ export class WebClientService {
...
+     */
+    public modifyConversation(conversation: threema.Conversation, isPinned?: boolean): Promise<null> {
+        const DATA_STARRED = 'isStarred';
+
+        // Prepare payload data
+        const args = {
+            [WebClientService.ARGUMENT_RECEIVER_TYPE]: conversation.type,
+            [WebClientService.ARGUMENT_RECEIVER_ID]: conversation.id,
+        };
+        const data = {};
```
lgrahl: `data` can be an empty field now. I guess that is okay but also a bit of unnecessary overhead.

src/partials/messenger.ts
dbrgn: Yep, but at least there is some feedback, versus none at all.

src/services/webclient.ts
lgrahl: You mean `undefined`. )Please don't use `null` any more. I've carefully removed all occurrences of `null`. :) )

src/partials/messenger.ts
lgrahl: My opinion is that we should either provide UI feedback along the line of *this particular thing didn't work* or we don't provide any UI feedback. It's helpful for the user to know that i.e. pinning a conversation didn't work but *oops, something didn't work but I can't tell you what* is not in any way helpful and would just confuse the user. Note that I would not consider showing a generic error an issue if there wasn't a noticeable delay between user action. But since we send messages back and forth, there is.

src/services/webclient.ts
dbrgn: I never said anything about `null` :slightly_smiling_face:

lgrahl: What was `None` about then? Nevermind. :man_shrugging:

src/partials/messenger.ts
dbrgn: If the user clicks the "pin" button, it's better if nothing happens and there's an "an error occurred" message 2 seconds later, than if simply nothing happens. The user will click again and again.

By showing a generic error message, this allows the user to actually describe something to support, which could then ask to take a look at the debug log. If we provide no feedback at all, the user will say "pinning does not work", which is very unclear.

But of course, if you think it's important we could create a specific error message for this condition and localize it, although it's unlikely that this will occur. There's not much that can go wrong when pinning (except bugs of course).

src/services/webclient.ts
dbrgn: Since this is not Python, "None" was meant in an English language kind of way, sorry if that was unclear. In the specification it's not rendered in monospace, maybe that's what confused you :slightly_smiling_face: To actually specify a null value, I'd use the msgpack `nil` type.

src/partials/messenger.ts
lgrahl: Mhh, that's a good point. The user will eventually figure out that this toast will have something to do with it furiously clicking the button. :grin: 

I wasn't referring to this particular rejection though but rather a question on whether we should do this for rejections in general. If we add toast messages for rejections, I think they should preferrably be specific and localised. That was my point.

dbrgn: FIne, I localized the error messages in the latest fixup commit.

public/img/ic_pin.svg
```diff
@@ -0,0 +1,10 @@
+<?xml version="1.0" encoding="UTF-8" standalone="no"?>
+<svg xmlns="http://www.w3.org/2000/svg"
+  height="24"
+  width="24"
+  viewBox="0 0 24 24">
+  <path
```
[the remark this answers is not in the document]
dbrgn: You're right. This is what it would look like at the same opacity:

screenshot

However, now it's lighter than the "back" arrow on the left and generally looks like it's disabled somehow. What do you think?

src/partials/messenger.conversation.html
dbrgn: :+1:

public/img/ic_pin.svg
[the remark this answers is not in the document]
lgrahl: It looks good to me.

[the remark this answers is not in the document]
dbrgn: Done.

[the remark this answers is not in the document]
dbrgn: Reverted it again, the "pinned" version looks bad with low contrast (it seems that the button is disabled).

dbrgn:
Applied changes. @lgrahl want to take another look?

src/partials/messenger.ts
```diff
@@ -503,6 +505,14 @@ class ConversationController {
                 .position('bottom center')
                 .hideDelay(hideDelayMs));
     }
+
+    public showMessage(msgTranslation: string, toastLength = 4000) {
```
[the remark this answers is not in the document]
dbrgn: Nice find :)

PR #610, opened 2018-10-10
Fix largeSingleEmoji setting

dbrgn:
It wasn't applied properly due to typos. TypeScript classes would have prevented that :slightly_smiling_face:  Unfortunately they don't work well as AngularJS directive controllers :slightly_frowning_face: 

Based on #601, merge that first. If you already want to review, only look at the last commit.


dbrgn:
Rebased.

PR #611, opened 2018-10-10
Confirmation acknowledgement fixes

lgrahl:
Make acknowledgements to `update/group`, `update/distributionList` and `update/contact` mandatory.


lgrahl:
What a silly question - of course not. :grin: On a more serious note, yeah, the three message types still need to be tested towards both Android and iOS just to be sure.

dbrgn:
Tested.

PR #613, opened 2018-10-11
Compose area: Properly insert newlines in Safari

### LINKED ISSUE #541: Linebreaks in Safari
(iOS Beta)

## Current Behavior

When using Safari, newlines (inserted with shift+enter) are not retained in sent messages.

dbrgn:
Fixes #541.


dbrgn:
The first line is a text node, not a div, thus we can handle all divs as separate lines.

Tested in Safari, Firefox and Chromium.

PR #617, opened 2018-10-12
Author of quotes message is too light

### LINKED ISSUE #616: Author of quotes message is too light
## Expected Behavior
Text should be darker.

## Current Behavior
Text is too light. This was changed in d14d73fef560650f40d92af9d3585e5db257c901.
screenshot 2018-10-12 at 07 56 57

## Possible Solution
Use the colors similar to the iOS app.
threema-20181210-080833-image

1. Quote a message in Threema Web
2. The text is very light

## Your Environment
* Threema Web version: 2.1.0-beta.6
* Threema App version: iOS 4.0 (2363)
* Browser name and version: Firefox 62.0.3
* Computer operating system and version: macOS 10.14
* Smartphone operating system and version: iOS 12.1 beta 3

joelfischerr:
Contact color now defaults to a color similar to what is used in iOS. Fixes #616.

screenshot 2018-10-12 at 08 51 57

### DIFF
--- a/src/directives/message_quote.ts
+++ b/src/directives/message_quote.ts
@@ -32,11 +32,6 @@ export default [
             controllerAs: 'ctrl',
             controller: [function() {
                 this.contact = () => webClientService.contacts.get(this.quote.identity);
-                this.contact.color = function() {
-                    if (isUndefined(this.contact.color)) {
-                        return '#2C65C9';
-                    }
-                };
             }],
             template: `
                 <div class="message-quote-content" ng-style="{'border-color': ctrl.contact().color}">
--- a/src/directives/message_quote.ts
+++ b/src/directives/message_quote.ts
@@ -34,7 +34,7 @@ export default [
                 this.contact = () => webClientService.contacts.get(this.quote.identity);
                 this.contact.color = function() {
                     if (isUndefined(this.contact.color)) {
-                        this.contact.color = '#2C65C9';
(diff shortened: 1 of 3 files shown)


dbrgn:
Thanks @joelfischerr! This will also influence the "default" color of avatars in the contact list before they are fetched though. Blue is a bit too "strong" for that.

If you want to create a proper workaround:

- In `threema.d.ts` make the `Receiver.color` attribute optional (add question mark).
- In `threema/container.ts`, remove the two `setDefault` lines for the color attribute
- In `directives/message_quote.ts` add a `color()` function to the controller. In that function, check whether a `color` attribute exists on the contact. If not, use your blue fallback color.
- Check for other usages of the `color` attribute and make sure that all of them are handled, now that the color is optional

If you don't have time / motivation for the change then let me know, so that someone else can implement it :slightly_smiling_face:

dbrgn:
@joelfischerr great! Yeah, that's the same issue (last bullet point). I'd also add a fallback color in that case.

joelfischerr:
This seems to be a bit above my level and I don't have enough time to get into it. I think it's better if someone else could implement it.

But for some reason just making 'Receiver.color' optional gives the desired results i.e. the design looks as it does in iOS (blue bar for quotes and black text for names).

dbrgn:
@joelfischerr no problem, thanks for trying! :slightly_smiling_face:

PR #620, opened 2018-10-16
Cancel timeouts on stop

### LINKED ISSUE #619: Chunks are sent in disconnected state
Here's a debug log:

```
[Debug] [StatusController] – "State change:" – "ok" – "->" – "error" (angular.js, line 15504)
[Debug] Sending thumbnail request for – "contact" – "z1f2qNrBA6U=" (angular.js, line 15504)
[Debug] Sending – "request/thumbnail" – "message" (angular.js, line 15504)
[Error] Cannot send task message in "closed" state
https://web-beta.threema.ch/node_modules/@saltyrtc/client/dist/saltyrtc-client.es5.js?v=2.1.0-beta.6:321
    (anonymous function) (angular.js:11814)
    (anonymous function) (angular.js:21558)
    completeTask (angular.js:21193)
    (anonymous function) (angular.js:6789)
[Error] Cannot send task message in "closed" state
https://web-beta.threema.ch/node_modules/@saltyrtc/client/dist/saltyrtc-client.es5.js?v=2.1.0-beta.6:321 – "Possibly unhandled rejection: {\"closeCode\":3001,\
    (anonymous function) (angular.js:11814)
    processChecks (angular.js:17943)
    $digest (angular.js:19074)
    $apply (angular.js:19462)
    (anonymous function) (angular.js:21563)
    completeTask (angular.js:21193)
    (anonymous function) (angular.js:6789)
```

Apparently while sending a message, the state changed.

The relevant code is in `sendChunk` (services/webclient.ts), a call to `this.relayedDataTask.sendMessage`. @lgrahl since the chunks should all be queued, can we simply log and disregard send errors?

dbrgn:
Fixes #619.


dbrgn:
@lgrahl ready for review!

src/services/timeout.ts
```diff
@@ -0,0 +1,86 @@
...
+        return timeout;
+    }
+
+    /**
+     * Cancel the specified timeout.
+     *
+     * Return true if the task hasn't executed yet and was successfully canceled.
+     */
+    public cancel<T>(timeout: ng.IPromise<T>): boolean {
+        this.logDebug('Cancelling timeout');
```
lgrahl: Naming which timeout would be great.

src/services/timeout.ts
```diff
@@ -0,0 +1,86 @@
...
+        if (this.config.DEBUG) {
+            this.$log.debug(this.logTag, msg);
+        }
+    }
+
+    /**
+     * Register a timeout.
+     */
+    public register<T>(fn: (...args: any[]) => T, delay: number, invokeApply: boolean, name?: string): ng.IPromise<T> {
+        this.logDebug('Registering timeout' + (name === undefined ? '' : ` (${name})`));
```
lgrahl: Forgot about string literals? :smiley_cat:

src/services/timeout.ts
```diff
@@ -0,0 +1,86 @@
...
+        this.config = config;
+        this.$log = $log;
+        this.$timeout = $timeout;
+    }
+
+    /**
+     * Log a message on debug log level, but only if the `DEBUG` flag is enabled.
+     */
+    private logDebug(msg: string): void {
+        if (this.config.DEBUG) {
```
lgrahl: I would drop that for now. This will be resolved by adding the micrologger from SaltyRTC.

src/services/timeout.ts
```diff
@@ -0,0 +1,86 @@
...
+        }
+    }
+
+    /**
+     * Register a timeout.
+     */
+    public register<T>(fn: (...args: any[]) => T, delay: number, invokeApply: boolean, name?: string): ng.IPromise<T> {
+        this.logDebug('Registering timeout' + (name === undefined ? '' : ` (${name})`));
+        const timeout = this.$timeout(fn, delay, invokeApply);
+        timeout.then(() => this.timeouts.delete(timeout)).catch((reason) => {
```
lgrahl: That is a bit awkward to read. I would suggest moving `.then` and `.catch` to separate lines.

src/services/timeout.ts
```diff
@@ -0,0 +1,86 @@
...
+            if (reason !== 'canceled') { // We can safely ignore cancellation
```
lgrahl: Is it really `canceled` with one `l`?

src/services/timeout.ts
```diff
@@ -0,0 +1,86 @@
...
+                this.$log.error(this.logTag, 'Promise rejected:', reason);
```
lgrahl: -> *Registered timeout promise rejected*?

src/services/timeout.ts
```diff
@@ -0,0 +1,86 @@
...
+        const cancelled = this.$timeout.cancel(timeout);
+        this.timeouts.delete(timeout);
+        return cancelled;
+    }
+
+    /**
+     * Cancel all pending timeouts.
+     */
+    public cancelAll() {
+        if (this.timeouts.size > 0) {
```
lgrahl: I would remove that conditional. It's perfectly fine to *cancel 0 timeouts* and I would rather get that information than nothing in the log.

src/services/webclient.ts
```diff
@@ -1120,6 +1126,9 @@ export class WebClientService {
             this.ackTimer = null;
         }
 
+        // Stop other timers
+        this.timeoutService.cancelAll();
```
lgrahl: This should only be done in case `close` is `true` (move it down). Furthermore, this also needs to be done when discarding a session after having tried to resume it.

TL;DR: Move this directly beside calls to `this.wireMessageFutures.clear()`.

src/services/timeout.ts
dbrgn: We don't know, as the promise does not contain the name. We could set the name as an attribute on the promise object, although that would be a bit ugly. What so you think?

lgrahl: IMO: Ugly but for debugging purposes okay.

src/services/timeout.ts
dbrgn: String literals? Do you mean template literals? I guess I could put the expression in one of those, but that would require nested template literals which is a bit ugly.

src/services/timeout.ts
dbrgn: The timeout logs are really noisy (dozens of logs when scrolling through the list of contacts), therefore I wouldn't do that.

src/services/timeout.ts
lgrahl: Sorry, yes, template literals. What about this:

```js
this.logDebug(`Registering timeout (${name === undefined ? 'unnamed' : name}`);
```

src/services/timeout.ts
dbrgn: Yes.

src/services/timeout.ts
lgrahl: Then please add a `TODO` note so we don't forget about it.

dbrgn: If we add a new logging wrapper, we'll have to touch every line of code that does logging anyways. The DEBUG condition is hard to overlooks :slightly_smiling_face:

dbrgn:
@lgrahl ok, good to merge now?

src/services/webclient.ts
```diff
@@ -1147,6 +1153,9 @@ export class WebClientService {
             // Remove all pending promises
             this.wireMessageFutures.clear();
 
+            // Stop pending timeouts
+            this.timeoutService.cancelAll();
```
lgrahl: Still missing at the other places where `this.wireMessageFutures.clear` is called. :)

dbrgn: Ah, that's what you meant :slightly_smiling_face:

dbrgn: Should be done now.

dbrgn:
All non-0 timeouts should now be converted to use the `TimeoutService`.

PR #622, opened 2018-10-16
Add more navigational landmarks and headings to improve screen reader…

### LINKED ISSUE #621: Add more navigational landmarks and headings to the UI
## Expected Behavior

Screen reader users should be able to move to more sections of the web UI via landmarks or heading navigation.

## Current Behavior

Except for the navigation landmark for the tabs, the UI is relatively flat, no way to easily find where the main contents starts after the messages list or the contacts, for example.

## Possible Solution

Adding a "main" landmark role and giving the conversation header text a heading, and also adding a blockquote role to quoted text in a message, and adding another heading for the start of unread messages makes the web UI much more navigable for screen reader users.

Pull request coming up.

1. Load Threema Web with a screen reader.
2. Try to easily find the start of the main conversation area.
3. Try to find the first unread message in a chat. Currently the only, and rather ineffective, way is by searching for the string "unread".
4. Try to discern in a quoted message what is the quote and what the reply.

## Your Environment

* Threema Web version: 2.1-beta6
* Threema App version: 4.0 latest TestFlight
* Browser name and version: Firefox 63
* Computer operating system and version: Windows 10
* Smartphone operating system and version: iOS 12.0.1

MarcoZehe:
… navigation. Fixes #621.

* Turn the section that contains the conversation into a main landmark.
* Make the header text that contains the contact or group name into a heading level 1.
* Make the text that denotes the start of unread messages into a heading level 3, to account for the possible heading level 2 below the header that sometimes appears.
* Make the quoted text in a message a blockquote (WAI-ARIA 1.2) so it is distinguishable from the reply.

This is all using non-intrusive WAI-ARIA roles and properties that adds screen reader semantics, but does not change the host language markup otherwise.

### DIFF
--- a/src/directives/message.html
+++ b/src/directives/message.html
@@ -64,7 +64,7 @@
     </div>
     <div ng-if="ctrl.message.statusType == 'firstUnreadMessage'" class="unread-separator">
         <div class="line"></div>
-        <p class="text" translate>messenger.UNREAD_MESSAGES</p>
+        <p class="text" role="heading" aria-level="3" translate>messenger.UNREAD_MESSAGES</p>
         <div class="line"></div>
     </div>
 </article>
--- a/src/directives/message_quote.ts
+++ b/src/directives/message_quote.ts
@@ -33,7 +33,7 @@ export default [
                 this.contact = () => webClientService.contacts.get(this.quote.identity);
             }],
             template: `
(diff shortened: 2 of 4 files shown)


dbrgn:
Great, thanks!

PR #624, opened 2018-10-16
StateService: Allow more time for loading initial data

dbrgn:
Now it takes 10.8s instead of 8.5s for the loading progress between 60% and 99% until the "slowConnect" flag is set.

This is because some people report that often the "slow connect" warning is shown for 1s before the data is finally loaded.

### DIFF
--- a/src/services/state.ts
+++ b/src/services/state.ts
@@ -206,7 +206,7 @@ export class StateService {
                 this.progress = 60;
                 this.progressInterval = this.$interval(() => {
                     if (this.progress < 80) {
-                        this.progress += 5;
+                        this.progress += 4;
                     } else if (this.progress < 90) {
                         this.progress += 2;
                     } else if (this.progress < 99) {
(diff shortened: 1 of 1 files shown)


dbrgn:
This is more of a hotfix. Here's an idea for a more granular loading indicator:

- In the loading screen, show a list of items that need to be loaded: Client info, Profile, Conversations, Contacts
- Next to each item is a loading indicator, or a checkmark
- Make the progress dependent on the number of initial data items loaded

If you think that's a good idea, I'll create a separate issue.

lgrahl:
For the upcoming issue:

> In the loading screen, show a list of items that need to be loaded: Client info, Profile, Conversations, Contacts. Next to each item is a loading indicator, or a checkmark. Make the progress dependent on the number of initial data items loaded.

Kind of reminds me of "The Sims" loading screens. :grin:
Fine with it but let's also factor in received chunks. We should add a small amount of statistics for that purpose (number of received chunks, last received chunk timestamps, number of sent chunks, last sent chunk, rx total, tx total, ...) which could then be used for several purposes, including the upcoming *send another push* change.

PR #629, opened 2018-10-17
Minimum delay futures

lgrahl:
Replaces the `ExecuteService`. Also, removal of dead code and reducing the minimum delay from 1s to 100ms.

Merge #620 first.


lgrahl:
Ok!

public/libs/future.js
```diff
@@ -20,6 +20,19 @@ class Future extends Promise {
...
+    static withMinDuration(promise, minDuration) {
+        const start = new Date();
+        return new Future((resolve, reject) => {
+            promise
+                .then((result) => {
+                    const timediff = new Date() - start;
+                    const delay = Math.max(minDuration - timediff, 0);
+                    self.setTimeout(() => resolve(result), delay);
+                })
+                .catch((error) => reject(error));
```
dbrgn: This line isn't necessary, right?

public/libs/future.js
```diff
@@ -20,6 +20,19 @@ class Future extends Promise {
         this._reject = reject;
     }
 
+    static withMinDuration(promise, minDuration) {
```
dbrgn: Maybe add a short docstring with a note/warning that this does not run inside the AngularJS loop.

PR #630, opened 2018-10-17
MarkupParser: Fix bug with double markup chars

dbrgn:
See test case.


lgrahl:
Perhaps a silly question but: What would happen when the same special character is used three times in a row? For example `***`. Maybe it makes sense to add a test case for that, too.

lgrahl:
Doesn't hurt. :)

PR #635, opened 2018-10-18
Make pin toggle button more accessible

dbrgn:
- Replace two-element button with single-element button
- Add `aria-pressed` attribute

It appears that by using `role="button"` and the `aria-pressed` attribute, browsers (at least Firefox and Safari) automatically bind the enter key event to the same handler as the click event. @MarcoZehe can we rely on that or would you always add an explicit keypress event handler?

### DIFF
--- a/src/components/toggle_button.ts
+++ b/src/components/toggle_button.ts
@@ -32,20 +32,20 @@ export default {
         iconEnabled: '@',
         iconDisabled: '@',
     },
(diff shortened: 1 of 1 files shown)


dbrgn:
Update: Key press event handlers should be added automatically by the `ngAria` module added in #636.

Closing in favor of #636 (cherry-picked the commit there).

PR #637, opened 2018-10-18
Clear typing flag when receiving message

dbrgn:


src/services/webclient.ts
```diff
@@ -2722,6 +2722,9 @@ export class WebClientService {
                     // Try to update it first. If not, add it as a new msg.
                     if (!this.messages.update(receiver, msg)) {
                         this.messages.addNewer(receiver, [msg]);
+
+                        // If we received a new message, then for sure the contact is not typing anymore.
```
lgrahl: Nit: wording. Suggestion:

> If we have received a new message, it is unlikely that the contact is still typing

PR #638, opened 2018-10-18
Fix updating of message caption

dbrgn:
The caption may change for example when receiving an image message on
Android. Because the internet connection may be slow, a create/message
will be sent initially (before downloading the image from the blob
server) without thumbnail and without caption (because the caption is
contained in the image EXIF data). Once the image is downloaded and
processed, an update will be sent with the caption.

### DIFF
--- a/src/directives/message_media.ts
+++ b/src/directives/message_media.ts
@@ -156,7 +156,7 @@ export default [
                         this.wasInView = inView;
 
                         if (!inView) {
-                            if (loadingThumbnailTimeout != null) {
+                            if (loadingThumbnailTimeout !== null) {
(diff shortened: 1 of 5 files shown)


src/directives/message_media.ts
```diff
@@ -156,7 +156,9 @@ export default [
                         this.wasInView = inView;
 
                         if (!inView) {
-                            timeoutService.cancel(loadingThumbnailTimeout);
+                            if (loadingThumbnailTimeout != null) {
```
lgrahl: `!==`

Also, where is this set to `null`?

dbrgn: It's the default.

PR #644, opened 2018-10-23
Use document focus instead of page visibility API

dbrgn:
For marking messages as unread and for triggering notifications, use
`document.hasFocus()` instead of the page visibility API.

This means that messages are only marked as read if the browser is in
the foreground and focussed. Otherwise, a notification is triggered.

Refs #211 and #17.


dbrgn:
Tested on Firefox/Chrome/Safari.

PR #653, opened 2018-10-25
More null safety

dbrgn:
Improve null correctness in controller models.

(Problems determined by temporarily enabling strict null checks.)

### DIFF
--- a/src/controller_model/contact.ts
+++ b/src/controller_model/contact.ts
@@ -22,6 +22,7 @@ import {AvatarControllerModel} from './avatar';
 import ControllerModelMode = threema.ControllerModelMode;
 
 export class ContactControllerModel implements threema.ControllerModel<threema.ContactReceiver> {
+    private logTag = '[ContactControllerModel]';
 
     // Angular services
     private $log: ng.ILogService;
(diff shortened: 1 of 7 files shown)


src/controller_model/contact.ts
```diff
@@ -49,37 +50,43 @@ export class ContactControllerModel implements threema.ControllerModel<threema.C
...
+        }
         this.webClientService = webClientService;
         this.mode = mode;
 
         switch (this.getMode()) {
             case ControllerModelMode.EDIT:
                 this.subject = $translate.instant('messenger.EDIT_RECEIVER');
-                this.firstName = this.contact.firstName;
-                this.lastName = this.contact.lastName;
+                this.firstName = this.contact!.firstName;
```
lgrahl: Might make sense to do `const contact = this.contact!` for these purposes but I don't have a strong opinion here.

dbrgn: Then TS complains about name shadowing of the `contact` variable. And calling the variable `editContact` is also ugly.

Let's keep the current approach for now :slightly_smiling_face:

Issue #652, opened 2018-10-25
Time format: Consider locale


dbrgn:
## Expected Behavior

Time format (24h vs am/pm) should consider the locale.

## Current Behavior

It [does not](https://github.com/threema-ch/threema-web/blob/dd515d1d06f0cbf62941ec9b55faca2b83990f98/src/filters.ts#L405-L408) :slightly_smiling_face:

PR #657, opened 2018-10-27
Refactor formatTime helper, fixes: #652

HVish:


dbrgn:
Hi, thanks for your pull request!

- The time includes the seconds, which we don't need. You can pass `{hour: '2-digit', minute:'2-digit'}` to the `options` to skip them. I think for now you could hardcode that in the `formatTime` helper function and remove the `options` parameter (which is unused so far).
- The tests need updating (see README for testing instructions). Unfortunately I found no way of mocking/overriding the currently defined locale for testing. But you could update the filter function so that the desired locale can be optionally passed in. The tests could then explicitly test both with `en-US` and `de-CH`.
- The linter complains about alphabetical sorting of imports. That one is easy, simply put the `formatTime` import before `filter`.
- The CLA needs to be signed for all contributions

dbrgn:
Closing due to inactivity.

PR #661, opened 2018-10-31
Contacts: Exclude hidden receivers

### LINKED ISSUE #655: Webclient shows contacts which are not visible in app
## Expected Behavior
When beeing invited to a group, members from this group are not automatically added to the local contact list in the Threema App. They also should not be visible in Threema Web.

## Current Behavior
The app is working as expected. The group members are not visible in the contact list in the app. But they are visible in Threema Web. 

## Steps to Reproduce
1. I was invited to a group chat (last year) with a lot of people I did not know before.
2. All these people have been added to my contact list in the app automatically (as this was the standard behaviour of the app months before).
3. Then there was an update of the app, which changed the mentioned behaviour: Contacts from groups will not be added to the contact list anymore, but existing group contacts stayed there.
4. I left the group (months ago) and deleted all the contacts of this group manually in the app. None of them have been linked to my adress book before and with none of them I had chats outside of the group.
5. When looking now at the contact list in Threema Web, all these old and deleted contacts are still visible there.

Note: I am not 100% sure if this bug is reproducible nowadays. Its a description of what happened and why I see more contacts in Threema Web than in the Threema App.
This bug report was created at advice of @dbrgn (https://threema-forum.de/thread-2908-post-40463.html#pid40463).

## Your Environment
* Threema Web version: 2.1.0
* Threema App version: 3.54.2000444 (Android via Google Play)
* Browser name and version: Firefox 62.0.3, 64 Bit
* Computer operating system and version: Linux Mint 19, Windows 7 Pro
* Smartphone operating system and version: LineageOS based on Android 7.1.2

dbrgn:
Contacts are marked as hidden if they are part of a common group, but have never been contacted directly.

Fixes #655.

### DIFF
--- a/src/filters.ts
+++ b/src/filters.ts
@@ -20,6 +20,7 @@ import {markify} from './markup_parser';
 import {MimeService} from './services/mime';
 import {NotificationService} from './services/notification';
 import {WebClientService} from './services/webclient';
+import {isContactReceiver} from './typeguards';
 
 angular.module('3ema.filters', [])
 
@@ -217,12 +218,27 @@ angular.module('3ema.filters', [])
 })
 
 /**
- * Return whether contact is not me.
+ * Filter receivers, exclude own contact.
  */
 .filter('isNotMe', ['WebClientService', function(webClientService: WebClientService) {
-    return function(obj: threema.Receiver) {
+    return function(receivers: threema.Receiver[]) {
         const valid = (contact: threema.Receiver) => contact.id !== webClientService.receivers.me.id;
-        return filter(obj, valid);
+        return filter(receivers, valid);
+    };
+}])
+
+/**
+ * Filter receivers, exclude hidden contact.
+ */
+.filter('isNotHidden', [function() {
+    return function(receivers: threema.Receiver) {
+        const valid = (receiver: threema.Receiver) => {
+            if (isContactReceiver(receiver)) {
+                return receiver.hidden !== true;
+            }
+            return true;
+        };
+        return filter(receivers, valid);
     };
 }])
 
--- a/src/filters.ts
+++ b/src/filters.ts
@@ -218,31 +218,6 @@ angular.module('3ema.filters', [])
 })
 
 /**
- * Filter receivers, exclude own contact.
- */
-.filter('isNotMe', ['WebClientService', function(webClientService: WebClientService) {
-    return function(receivers: threema.Receiver[]) {
-        const valid = (contact: threema.Receiver) => contact.id !== webClientService.receivers.me.id;
-        return filter(receivers, valid);
-    };
-}])
-
-/**
- * Filter receivers, exclude hidden contact.
- */
-.filter('isNotHidden', [function() {
-    return function(receivers: threema.Receiver) {
-        const valid = (receiver: threema.Receiver) => {
-            if (isContactReceiver(receiver)) {
-                return receiver.hidden !== true;
-            }
-            return true;
-        };
-        return filter(receivers, valid);
-    };
-}])
-
-/**
  * Filter for duration formatting.
  */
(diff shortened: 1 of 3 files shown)


src/partials/messenger.navigation.html
```diff
@@ -120,7 +120,7 @@
 <div id="navigation-contacts" class="tab-content" ng-if="ctrl.activeTab == 'contacts'" in-view-container>
     <p class="empty" ng-if="ctrl.contacts().length === 0" translate>messenger.NO_CONTACTS_FOUND</p>
     <ul ng-class="{'hide-inactive': ctrl.hideInactiveContacts()}">
-        <li ng-repeat="contact in ctrl.contacts() | isNotMe | filter:ctrl.searchContact"
+        <li ng-repeat="contact in ctrl.contacts() | isNotMe | isNotHidden | filter:ctrl.searchContact"
```
lgrahl: IIRC all of these will run their own filter loop? Wouldn't it make sense to create a combined filter where conditionals can be added in form of a callback?

Something like: `filter(isNotMe, isNotHidden, ...)`

src/filters.ts
```diff
@@ -217,12 +218,27 @@ angular.module('3ema.filters', [])
...
-        return filter(obj, valid);
+        return filter(receivers, valid);
+    };
+}])
+
+/**
+ * Filter receivers, exclude hidden contact.
+ */
+.filter('isNotHidden', [function() {
+    return function(receivers: threema.Receiver) {
```
lgrahl: (Reply to https://github.com/threema-ch/threema-web/pull/661/files#r229691268)

This is a function that accepts some kind of iterable. `filter` accepts a list or an object and filters it with the given callback. Thus, unless there is some magic going on, each filter will go through the whole list.

dbrgn: Ah *that's* what you mean :slightly_smiling_face: You are right, sorry. I'll fix!

src/partials/messenger.navigation.html
dbrgn: Resolved in the other conversation. @lgrahl was right :slightly_smiling_face:

src/filters.ts
dbrgn: Fixed in the fixup commit. I decided to ditch the filters altogether and to reuse the iterator we already have.

lgrahl: Yeah, I think this filter is ancient and IIRC has been used in other places previously.

src/partials/messenger.ts
```diff
@@ -942,7 +942,24 @@ class NavigationController {
     }
 
     public contacts(): threema.ContactReceiver[] {
-        return Array.from(this.webClientService.contacts.values()) as threema.ContactReceiver[];
+        const contacts = [];
+        // Since `values()` on a map returns an iterator, we cannot directly
+        // apply the `.filter()` method that arrays provide. Therefore, in
+        // order to avoid creating an intermediate array just for filtering,
+        // create the list of contacts imperatively.
+        for (const contact of this.webClientService.contacts.values()) {
```
lgrahl: Why not .filter()?

dbrgn: You mean on the map directly? Would be possible, but on the other hand we never need the keys for filtering (and `values()` ignores those completely).

If you meant `.values().filter()`, then see the comment above :slightly_smiling_face:

lgrahl: Oh, I completely missed that. We *could* use generators and `yield` but since we only want an array as a result this would probably be overkill.

dbrgn: TIL that generators exist in JS... https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Statements/function* But yes, there's no advantage since we'll convert to an array anyways.

PR #669, opened 2018-11-14
Replace Test Cafe with Selenium

dbrgn:
The reason for this is that testcafe cannot properly simulate some keystrokes like SHIFT+ENTER.

Selenium requires some binary components, but that is OK since they can be installed via npm.

I had to write a custom test runner for this, but I quite like the end result :slightly_smiling_face:

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -22,28 +22,28 @@ jobs:
     steps: *test-steps
     environment:
       BUILDTARGET: build:unittests
-      TESTTARGET: test:unittests
+      TESTTARGET: test:unit
   unittest-node10:
     docker:
       - image: circleci/node:10-browsers
     steps: *test-steps
     environment:
       BUILDTARGET: build:unittests
-      TESTTARGET: test:unittests
+      TESTTARGET: test:unit
   uitest-firefox:
     docker:
       - image: circleci/node:10-browsers
     steps: *test-steps
     environment:
       BUILDTARGET: build:uitests
-      TESTTARGET: test:uitests firefox:headless
+      TESTTARGET: test:ui firefox
   uitest-chrome:
     docker:
       - image: circleci/node:10-browsers
     steps: *test-steps
     environment:
       BUILDTARGET: build:uitests
-      TESTTARGET: test:uitests firefox:headless
+      TESTTARGET: test:ui chrome
   lint:
     docker:
       - image: circleci/node:8-browsers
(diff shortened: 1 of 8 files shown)


.circleci/config.yml
```diff
@@ -36,14 +36,14 @@ jobs:
...
       BUILDTARGET: build:uitests
-      TESTTARGET: test:uitests firefox:headless
+      TESTTARGET: test:uitests firefox
   uitest-chrome:
     docker:
       - image: circleci/node:10-browsers
     steps: *test-steps
     environment:
       BUILDTARGET: build:uitests
-      TESTTARGET: test:uitests firefox:headless
```
dbrgn: Oops :smile:

dbrgn:
Successfully tested on:

- Linux / Firefox
- Linux / Chromium
- macOS / Firefox
- macOS / Safari

.circleci/config.yml
lgrahl: Who needs Chrome :grin:

tests/ui/run.ts
```diff
@@ -3,59 +3,134 @@
...
+    let success = 0;
+    let failed = 0;
+    console.info('\n====== THREEMA WEB UI TESTS ======\n');
+    try {
+        for (const [name, testfunc] of TESTS) {
+            console.info(TermColor.blue(`» ${i + 1}: Running test: ${name}`));
+            await driver.get(TEST_URL);
+            await testfunc(driver);
+            success++;
+            i++;
```
lgrahl: I don't see how this will work in case a test fails. IIRC that would break the loop early.

dbrgn: Yes, currently the test runner does an early abort. I think that's fine, although we could change the logic so that other tests are run even if there was a failure.

tests/ui/run.ts
```diff
@@ -3,59 +3,134 @@
...
+        }
+    } catch (e) {
+        console.error(TermColor.red(`\nTest failed:`));
+        console.error(e);
+        failed++;
+    } finally {
+        await driver.quit();
+    }
+    const colorFunc = failed > 0 ? TermColor.red : TermColor.green;
+    console.info(colorFunc(`\nSummary: ${i} tests run, ${success} succeeded, ${failed} failed`));
```
lgrahl: I think I was confused about this debug entry... `i` is logged but not the total amount of tests.

dbrgn: Yes, as the text says, it prints how many tests were run, not how many there are. If there are no failures, then all tests will be run.

lgrahl: The point being: The code looks like it supports multiple failures while in reality only supports one.

Issue #674, opened 2018-11-19
Compose Area: Cursor position wrong after inserting emoji


dbrgn:
Steps to reproduce (Firefox):

Put cursor between two lines:

![image](https://user-images.githubusercontent.com/105168/48703982-ad313c00-ebf5-11e8-8e93-f44ea25a3b56.png)

Insert emoji:

![image](https://user-images.githubusercontent.com/105168/48703986-b15d5980-ebf5-11e8-83f6-84399a93e74f.png)

Then type some text. The text will be on the next line, not next to the emoji.

PR #675, opened 2018-11-19
ComposeArea: Fix cursor pos when inserting emoji

dbrgn:
Another one fixed!

See #674.


ovalseven8:
Obsolete?

dbrgn:
> Obsolete?

Yes, thank you. Replaced by #773.

PR #689, opened 2018-12-12
Apply new scrollbar styling properties

dbrgn:
Firefox 64+ now supports styling scrollbars.

https://developer.mozilla.org/en-US/docs/Web/CSS/CSS_Scrollbars
https://drafts.csswg.org/css-scrollbars-1/

Since it doesn't support setting an absolute width, I set the width to
`auto` and also increased the width for webkit based browsers a bit.

Additionally, I increased the contrast a bit.

### DIFF
--- a/src/sass/components/_scrollbars.scss
+++ b/src/sass/components/_scrollbars.scss
@@ -1,10 +1,18 @@
-// TODO: Prefixes
+// Webkit specific global styling
 
 ::-webkit-scrollbar {
-    width: 8px;
-    background-color: #fcfcfc;
+    width: 6px;
+    background-color: $scrollbar-color-bg;
     border: 1px solid #f9f9f9;
 }
+
 ::-webkit-scrollbar-thumb {
-    background-color: #ddd;
+    background-color: $scrollbar-color-thumb;
+}
+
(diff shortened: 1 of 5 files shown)


src/sass/components/_scrollbars.scss
```diff
@@ -1,10 +1,15 @@
 // TODO: Prefixes
 
 ::-webkit-scrollbar {
-    width: 8px;
-    background-color: #fcfcfc;
+    width: 10px;
```
lgrahl: If we go for thin, I'd say `6px`.

src/sass/components/_scrollbars.scss
```diff
@@ -1,10 +1,15 @@
...
+    background-color: $scrollbar-color-bg;
     border: 1px solid #f9f9f9;
 }
 ::-webkit-scrollbar-thumb {
-    background-color: #ddd;
+    background-color: $scrollbar-color-thumb;
+}
+
+@mixin scrollbar {
+    scrollbar-width: auto;  // Unfortunately Firefox does not support exact pixel values
```
lgrahl: Voting for `thin`. :slightly_smiling_face:

dbrgn: Done.

PR #705, opened 2019-01-07
Fix budo config

dbrgn:
Budo was not set up properly. Replaces #704.

This does not work properly yet and is blocked by https://github.com/mattdesl/budo/issues/239.


dbrgn:
Merging a workaround for now. Note that this does not include the preset-env options when running the devserver.

PR #706, opened 2019-01-07
Fix collapsing of consecutive whitespace in message

### LINKED ISSUE #703: On pasting text, multiple whitespace characters are collapsed
This is only a display issue though, the value sent will contain the proper whitespace.

I believe the reason is that the innerHTML is set to contain multiple space characters, which are collapsed according to HTML display rules.

We should replace consecutive whitespaces with the unicode equivalent or with `&nbsp;` when pasting.

dbrgn:
This should fix #703.

Unfortunately I didn't find out how to create a Selenium test for this, since white-space collapsing is only display logic and will not be reflected in the DOM.

### DIFF
--- a/src/sass/sections/_compose_area.scss
+++ b/src/sass/sections/_compose_area.scss
@@ -40,6 +40,7 @@ compose-area {
                     max-height: calc(1.3em * 5);
                     min-height: 22px;
                     cursor: text;
+                    white-space: pre-wrap;
 
                     img.e1 {
                         vertical-align: middle;
--- a/src/sass/sections/_conversation.scss
+++ b/src/sass/sections/_conversation.scss
@@ -351,6 +351,10 @@
                 @include mouse-hand;
             }
 
+            span {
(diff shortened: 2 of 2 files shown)


lgrahl:
Some test failed, though.

dbrgn:
Yes, in Chrome :cry: Also happens locally.

I can't explain why the display logic would affect the DOM content. I'm not sure how contenteditable could be any more broken...

dbrgn:
When entering "a<shift+enter>b<shift+enter>c<shift+enter>", Chromium does this:

1. With `white-space: pre-wrap`:

img

2. Without `white-space: pre-wrap`:

img

---

When this is followed by inserting an emoji, Chromium does this:

With `white-space: pre-wrap`:

img

Without `white-space: pre-wrap`:

img

 (╯°□°）╯︵ ┻━┻

dbrgn:
:cry: At least tests pass now.

PR #715, opened 2019-01-16
Fix out-of-sync bug on Android reconnect

### LINKED ISSUE #712: Out of sync on reconnect

1. Open a conversation
2. Type a message, don't send it yet
3. Trigger a background reconnect
4. After the connection is back but before the conversations have been loaded again, send the message

The message is sent, but not visible in the conversation.

dbrgn:
This is work in progress. It's a bit tricky.

When Android reconnects, the messages and the conversations are reloaded. If a user submits a message after the receivers are already loaded but before the conversations are loaded, the unsent message gets put into a new(?) conversation which is then overwritten by the real conversations, or something like that.

### DIFF
--- a/src/partials/messenger.ts
+++ b/src/partials/messenger.ts
@@ -521,7 +521,10 @@ class ConversationController {
     public submit = (type: threema.MessageContentType, contents: threema.MessageData[]): Promise<any> => {
         // Validate whether a connection is available
         return new Promise((resolve, reject) => {
-            if (!this.stateService.readyToSubmit(this.webClientService.chosenTask)) {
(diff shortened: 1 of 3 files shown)


dbrgn:
@rugk please don't misuse the review system for comments that don't add to the discussion. This is clearly WIP.

rugk:
Ah, so these comments are WIP okay then. Missed the WIP a bit, so I thought it's a PR, so it's open for review…

lgrahl:
If it's not WIP any more, you may want to edit your first message. :slightly_smiling_face:

PR #719, opened 2019-01-21
Handle unknown contacts when checking feature mask

### LINKED ISSUE #718: Cannot send pictures to a group
A customer could not send pictures to a group:

img

> TypeError: Cannot read property 'displayName' of undefined

The error message is probably a red herring, I'd think that it's probably a feature mask issue.

dbrgn:
When sending an image to a group, if a group contact cannot be found,
then that would result in the contact being `undefined`.

Thus, in that case we may not access the `displayName` attribute.

Hopefully fixes #718.

### DIFF
--- a/src/services/webclient.ts
+++ b/src/services/webclient.ts
@@ -1612,8 +1612,7 @@ export class WebClientService {
                             if (this.mimeService.isAudio(mime)) {
                                 requiredFeature = ContactReceiverFeature.AUDIO;
                                 invalidFeatureMessage = 'error.AUDIO_MESSAGES_NOT_SUPPORTED';
-                            } else if (this.mimeService.isImage(mime)
-                                || this.mimeService.isVideo(mime)) {
+                            } else if (this.mimeService.isImage(mime) || this.mimeService.isVideo(mime)) {
                                 requiredFeature = ContactReceiverFeature.AUDIO;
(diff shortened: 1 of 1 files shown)


src/services/webclient.ts
```diff
@@ -1637,16 +1637,18 @@ export class WebClientService {
                                     if (identity !== this.me.id) {
                                         // tslint:disable-next-line: no-shadowed-variable
                                         const contact = this.contacts.get(identity);
-                                        if (contact === undefined
-                                            || !hasFeature(contact, requiredFeature, this.$log)) {
+                                        if (contact === undefined) {
+                                            this.$log.error(`Cannot retrieve contact ${identity}`);
+                                        } else if (!hasFeature(contact, requiredFeature, this.$log)) {
```
dbrgn: This is the main change. The rest is just code formatting.

lgrahl: is it okay that an undefined contact is now a *supported member*?

lgrahl: In that case, can you add a note why you've made the change? Otherwise if I'd read that code in a few months time...

lgrahl: I like comments for code sections that are really weird and I consider this to be one of the considerably weird conditions. :wink:

Issue #716, opened 2019-01-17
Feature: indicate connection status in favicon


caugner:
## Expected Behavior
If Threema Web is not connected, the favicon should indicate this with a red circle/dot.

## Current Behavior
The favicon is always the same.

## Possible Solution
Change the favicon dynamically (using SVG?).
See also #355.

Issue #355, opened 2017-10-02
Feature: change favicon.ico for unread messages


fichtennadel:
Changing the favicon.ico if there are unread messages would be nice, for example add a red marker.
![image](https://user-images.githubusercontent.com/2324670/31099421-37d13da8-a7c6-11e7-9fae-3191c50926c7.png)

Change back to normal, when messages are read.

(idea from Slack, quite nice in bookmark toolbar in FF)

PR #720, opened 2019-01-21
Show error connection state in favicon

dbrgn:
Experimental solution for #716.

It might not work in all browsers (see the load of different favicon formats in `index.html`) but it seems to work in Firefox at least.

img

The red dot is shown if the global connection state is set to "error" (red top bar).

Opinions?


lgrahl:
I don't particularly like how the red dot overlays one of the green ones. Perhaps moving it to the top right might be a better fit. Apart from that, any chance we can solve #355 easily along the line?

dbrgn:
Any suggestion on how to integrate #355 visually? Maybe a green dot?

dbrgn:
Suggestion:

favicon-16x16
favicon-16x16-offline
favicon-16x16-unread
favicon-32x32
favicon-32x32-offline
favicon-32x32-unread

(Those variants are now generated with a script, so we can adjust colors and dimensions if desired.)

ovalseven8:
Mh, I don't really like those icons. They're not nice and don't indicate their meaning. :/

lgrahl:
No disrespect but that comment isn't really helpful. Suggestions?

ovalseven8:
> No disrespect but that comment isn't really helpful. Suggestions?

Well, you're right. In the start post @dbrgn asked for opinions though. Thing is, the icons don't indicate any meaning, at least for me. Perhaps the designers at Threema can come up with suggestions?

What could look okay is a small circular border around the Threema icon which switches colors.

lgrahl:
The border is already very thin (for an icon), so I'm not sure this would be a good idea.

joelfischerr:
Doesn't the red badge at the top right look like the notification badge on iOS and macOS?
While the notification badge could mean that an error occurred, it usually is just a notification. I don't think many users would be able to recognise it as an error on first glance or even at all.

Would a red circle with an exclamation mark in the middle work as the icon for an error?

dbrgn:
> Would a red circle with an exclamation mark in the middle work as the icon for an error?

While that would probably work for the larger variant, it will not be visible at 16x16px.

Or did you mean replacing the entire icon with just a red circle? Then it's not clear that it's the Threema Web tab anymore.

(The meaning of the examples above would be red dot = error, blue dot = unread messages available.)

dbrgn:
We decided to go ahead with the colored circles for now. I also added a yellow version that reflects the "warning" state.

favicon-16x16 OK
favicon-16x16-offline Connecting or reconnecting
favicon-16x16-warning Connection error
favicon-16x16-unread Unread messages

Support for the unread message indicator is now implemented.

For iOS devices the warning favicon is not used, since being disconnected frequently is expected.

PR #732, opened 2019-01-24
Update README.md

ovalseven8:
:D 

What happens with god old security"at"threema.ch? 😄


dbrgn:
The 2019 address is not currently in use (yet?).

We moved to the suffixed version due to a lot of incoming spam.

ovalseven8:
I've created this pull request because that's the email address listed at the official website:
https://threema.ch/en/contact

PR #745, opened 2019-01-29
Replace U+00A0 with real spaces before sending text

### LINKED ISSUE #731: On pasting text, spaces are replaced with &nbsp;
Unfortunately, the hacky workaround introduced in #706 causes some problems. I expected non breaking spaces to be sent as regular spaces, but of course that's not the case.

Only leading space should be replaced with non-breaking space for now. The proper solution would of course be a CSS fix, this is blocked by the compose area rewrite though.

dbrgn:
This is a workaround for the problems introduced in #706. It's not nice, but the hack can be removed once we have the new compose area.

Note that when pasting large texts from the clipboard, it will still look strange in the compose area itself (since non-breaking spaces don't break :wink:) but the submitted text will not contain non-breaking spaces anymore (which caused display issues in the apps themselves).

Fixes #731.

### DIFF
--- a/src/directives/compose_area.ts
+++ b/src/directives/compose_area.ts
@@ -170,7 +170,16 @@ export default [
                 //
                 // Emoji images are converted to their alt text in this process.
                 function submitText(): Promise<any> {
-                    const text = extractText(composeDiv[0], logAdapter($log.warn, logTag));
+                    const rawText = extractText(composeDiv[0], logAdapter($log.warn, logTag));
+
+                    // Due to #731, and the hack introduced in #706, the
+                    // extracted text may contain non-breaing spaces (U+00A0).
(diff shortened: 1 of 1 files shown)


src/directives/compose_area.ts
```diff
@@ -170,7 +170,16 @@ export default [
                 //
                 // Emoji images are converted to their alt text in this process.
                 function submitText(): Promise<any> {
-                    const text = extractText(composeDiv[0], logAdapter($log.warn, logTag));
+                    const rawText = extractText(composeDiv[0], logAdapter($log.warn, logTag));
+
+                    // Due to #731, and the hack introduced in #706, the
+                    // extracted text may contain non-breaing spaces (U+00A0).
```
lgrahl: Typo. But leave it as we want this removed anyway. :grin:

PR #761, opened 2019-02-27
fix 'sed -i' on BSD

joelfischerr:
The `-i.bak` notation is required so that the sed command works both on Linux and on macOS: https://stackoverflow.com/q/5694228/284318

### DIFF
--- a/dist/package.sh
+++ b/dist/package.sh
@@ -83,11 +83,11 @@ for target in "${targets[@]}"; do
     else
         install -D "node_modules/$target" "$DIR/node_modules/$target"
     fi
-    sed -i "/sourceMappingURL/d" "$DIR/node_modules/$target"
+    sed -i.bak "/sourceMappingURL/d" "$DIR/node_modules/$target"
 done
 
 echo "+ Update version number..."
-# Note: The `-i.bak` notation is requires so that the sed command works both on Linux
+# Note: The `-i.bak` notation is required so that the sed command works both on Linux
 # and on macOS: https://stackoverflow.com/q/5694228/284318
 sed -i.bak -e "s/\[\[VERSION\]\]/${VERSION}/g" $DIR/index.html $DIR/troubleshoot/index.html $DIR/dist/app.js $DIR/manifest.webmanifest $DIR/browserconfig.xml $DIR/version.txt
 rm $DIR/index.html.bak $DIR/troubleshoot/index.html.bak $DIR/dist/app.js.bak $DIR/manifest.webmanifest.bak $DIR/browserconfig.xml.bak $DIR/version.txt.bak
--- a/dist/package.sh
+++ b/dist/package.sh
@@ -83,12 +83,12 @@ for target in "${targets[@]}"; do
     else
         install -D "node_modules/$target" "$DIR/node_modules/$target"
     fi
+    # Note: The `-i.bak` notation is required so that the sed command works both on Linux
+    # and on macOS: https://stackoverflow.com/q/5694228/284318
     sed -i.bak "/sourceMappingURL/d" "$DIR/node_modules/$target"
 done
 
 echo "+ Update version number..."
-# Note: The `-i.bak` notation is required so that the sed command works both on Linux
-# and on macOS: https://stackoverflow.com/q/5694228/284318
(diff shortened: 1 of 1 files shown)


dist/package.sh
```diff
@@ -83,11 +83,11 @@ for target in "${targets[@]}"; do
...
         install -D "node_modules/$target" "$DIR/node_modules/$target"
     fi
-    sed -i "/sourceMappingURL/d" "$DIR/node_modules/$target"
+    sed -i.bak "/sourceMappingURL/d" "$DIR/node_modules/$target"
 done
 
 echo "+ Update version number..."
-# Note: The `-i.bak` notation is requires so that the sed command works both on Linux
+# Note: The `-i.bak` notation is required so that the sed command works both on Linux
 # and on macOS: https://stackoverflow.com/q/5694228/284318
```
lgrahl: And since we use `sed` multiple times, I'd like this comment updated and moved to the top one (or just removed altogether).

dist/package.sh
```diff
@@ -83,11 +83,11 @@ for target in "${targets[@]}"; do
     else
         install -D "node_modules/$target" "$DIR/node_modules/$target"
     fi
-    sed -i "/sourceMappingURL/d" "$DIR/node_modules/$target"
+    sed -i.bak "/sourceMappingURL/d" "$DIR/node_modules/$target"
```
lgrahl:
```suggestion
    sed -i"" "/sourceMappingURL/d" "$DIR/node_modules/$target"
```

dist/package.sh
```diff
@@ -83,11 +83,11 @@ for target in "${targets[@]}"; do
...
 done
 
 echo "+ Update version number..."
-# Note: The `-i.bak` notation is requires so that the sed command works both on Linux
+# Note: The `-i.bak` notation is required so that the sed command works both on Linux
```
lgrahl:
```suggestion
# Note: The `-i""` notation is required so that the sed command works both on Linux
```

dist/package.sh
```diff
@@ -83,11 +83,11 @@ for target in "${targets[@]}"; do
...
 # and on macOS: https://stackoverflow.com/q/5694228/284318
 sed -i.bak -e "s/\[\[VERSION\]\]/${VERSION}/g" $DIR/index.html $DIR/troubleshoot/index.html $DIR/dist/app.js $DIR/manifest.webmanifest $DIR/browserconfig.xml $DIR/version.txt
```
lgrahl:
```suggestion
sed -i"" -e "s/\[\[VERSION\]\]/${VERSION}/g" $DIR/index.html $DIR/troubleshoot/index.html $DIR/dist/app.js $DIR/manifest.webmanifest $DIR/browserconfig.xml $DIR/version.txt
```

dist/package.sh
```diff
@@ -83,11 +83,11 @@ for target in "${targets[@]}"; do
...
 rm $DIR/index.html.bak $DIR/troubleshoot/index.html.bak $DIR/dist/app.js.bak $DIR/manifest.webmanifest.bak $DIR/browserconfig.xml.bak $DIR/version.txt.bak
```
lgrahl:
```suggestion
```

dist/package.sh
joelfischerr: afaik `sed -i""` doesn't work on BSD/macOS. But unless we can safely delete all `.bak` files in `
"$DIR/node_modules/$target"`, it might be better to let everyone using macos (which is currently only me anyways) install gnu-sed.

dist/package.sh
joelfischerr: I removed the comment.

dist/package.sh
lgrahl: Tested and works fine on Mac OS

lgrahl: Actually... no, it creates a file suffixed with `-e`. :slightly_frowning_face:

joelfischerr:
sorry for the mess 🙈

dbrgn:
For more context, see https://github.com/threema-ch/threema-web/commit/bf16b248f68746797fec36b8ef5e25a799498275#diff-f5391d2d698915c0fba5dca0f0709661 and https://github.com/threema-ch/threema-web/commit/7f0c47880e85dd0e5e172cf0bde4a5f54f33568c#diff-f5391d2d698915c0fba5dca0f0709661.

`sed -i""` will not work on Linux iirc.

dbrgn:
img

lgrahl:
Nah, I need to apologise. Looks like we should go back to your initial commit (you can force-push it) and then only move the comment to the top `sed` call. :sweat_smile:

joelfischerr:
Don't worry. This should now work right?

Issue #9, opened 2017-02-15
Dark theme setting is not available


xf-:
Threema looks unfamiliar with a white theme. Synchronize the theme.

rugk:
But preferably make it possible to switch the setting on the desktop independently from the app. Users may want to use a dark Threema App and a white Threema Web.

xf-:
@rugk i thought the same, but there is no settings area available in web. It needs to be stored on the phone for better usability with multiple devices or outdated sessions.
First step is synchronization easier, but a setting would be great :)

rugk:
I think storing this setting could be done in different ways:
* locally in the browser (aka local storage), possibly encrypted with the password
* on the device (thus the setting would be the same for every web client connected)
* on a new connection default to the setting of the app and show switch in the web client for (temporarily) switching the design

The first and last thing can of course be combined, so that permanent session store the setting while temporary session always use the one from the app.

FloThinksPi:
Hey guys, i worked some basic stuff out and Implemented a prototype over the local Storage. Tell me what you think about it. https://github.com/FloThinksPi/threema-web.
It does not apply the Themes on every element atm, i guess HTML and CSS need some tags but i have no time left today to look into that. Also the changes get not applyed directly on the Settings Dialog but yea :yum:

dbrgn:
@FloThinksPi cool! If you create a pull request (can also be marked as work in progress) we can give you feedback on it :)

Regarding storage of the settings:

- Local storage is generally a good idea.
- We could store the settings together with the keys (encrypted with the password), but that would mean that the settings are only persisted when using stored sessions.
- Storing them separately, settings can be persisted independent of the session. That would also mean that synchronizing web theme and app theme is tricky though, because it's unclear when to override user defined settings.

Maybe @sillych has an opinion on that.

rugk:
IMHO a good solution would be to sync the setting for temporary sessions and offer a switch in the web UI to change the theme. This switch should however not save any data and only apply to the current session.
When using a persistent session the setting can be changed encrypted in local storage. In this case it would not be synced with the app.

Or you just do it as you said and store it unencrypted in local storage (not nice, but not a problem since the data is really not useful to an attacker) and do not sync it with the app.

IMHO it's your decision.

FloThinksPi:
Hey @rugk and @dbrgn, thansk for the response and ideas.

In my Opinion users often considder the Webapp as standalone application and dont know about how strongly connected The actuall app and the webapp are.
[...]

PR #763, opened 2019-02-28
[WIP] Dark Theme

joelfischerr:
Should solve: [Dark theme setting is not available #9](https://github.com/threema-ch/threema-web/issues/9)

To manage expectations, this is the first time I have tried to contribute code to anything that is bigger than a few hundred lines. So things are probably a bit (very) rough.
If you can point me to the general direction of the changes needed, I'm eager to get it up to the standard of the rest of the code.

I have basically tried to do what @dbrgn outlined in https://github.com/threema-ch/threema-web/issues/9#issuecomment-398373861.

## Todo:
- [ ] Clean up the colors in `app-dark/light.scss`
- [ ] Improve loading the css files (https://github.com/threema-ch/threema-web/pull/763#issuecomment-471542622)
- [ ] Remove the remaining differences between the current theme and the new light theme
- [x] The placeholder in the send file/image dialog is not themed (all other placeholders are, I don't yet know what is wrong here)
- [x] Check lint disable in `_dialogs.scss`
- [x] Poll and image icons (shown in message preview in navigation) are not yet themed as they are image files (light and dark options are probably needed)
- [x] Loading indicator is not yet themed as it is a gif
- [x] Message quote colors are incorrect, I don't yet know where these colors are coming from.
- [x] The color of `ic_pin.svg` is not updated on theme change.


dbrgn:
@joelfischerr what's the status on this PR? Is it ready to review? Or are there still many open issues?

dbrgn:
Thanks @joelfischerr! I looked over the changes and tried it, and it looks great so far.

There are certainly things that can be improved (e.g. the way the theme is loaded), but that can wait until the detail work on the styles is finished.

So for now, I think the approach is good and the next steps would be to ensure that the light theme stays the same as before and to finish the changes required in the dark theme.

If there is something you need from me (e.g. an updated loading spinner GIF), let me know!

joelfischerr:
I have not quite finished replicating the original theme, but I think I'm making progress.

There are some custom icons that are loaded as images. Changing the colors in those is quite easy but I'm not sure how to load them when the theme changes. What do you think would be the best way?

dbrgn:
Which icons are you talking about?

joelfischerr:
I think all files starting with `ic_` in `public/img/` are loaded as images while the vector icons from Angular Material are inlined allowing them to be themed with CSS ([material.angular.io/components/icon/overview#svg-icons](https://material.angular.io/components/icon/overview#svg-icons)).

Issue #751, opened 2019-02-13
Too long quoted message destroys layout


rugk:
## Expected Behavior

Try to break as few things as possible…

## Current Behavior

It overflows the input field. After switching the chat back and forth, it now bumped all the chat window content to the top (header is invisible e.g.), but I can at least enter text.

Everything is back/okay after sending that message.

## Steps to Reproduce (for bugs)

1. Send a message with many many lines of text…
2. Quote it.

## Your Environment

* Threema Web version: 2.1.7-gh
* Browser name and version: Firefox 67

joelfischerr:
I think have a quick fix ready at: https://github.com/joelfischerr/threema-web/commit/00499118eb3d352d64addaa41092b19f0104a460.
It doesn't show an ellipsis at the end like the iOS version does, but I don't think showing it is possible without a lot more work. If the Android app shows an ellipsis as well it might be worth it though.

PR #767, opened 2019-03-01
Line clamping when quoting text

joelfischerr:
Should fix: #751 Too long quoted message destroys layout

It doesn't show an ellipsis (which iOS currently does) at the end which is apparently not easy (https://css-tricks.com/line-clampin/).

Used to look like:
old
Now looks like:
new

### DIFF
--- a/src/sass/sections/_conversation.scss
+++ b/src/sass/sections/_conversation.scss
@@ -65,7 +65,6 @@
             max-width: 85%;
             max-height: 9.4em;
             overflow: hidden;
-            text-overflow: ellipsis;
             line-height: 1.3em;
         }
     }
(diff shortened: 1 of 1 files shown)


src/sass/sections/_conversation.scss
```diff
@@ -63,6 +63,9 @@
         .message-quote {
             @include word-wrap;
             max-width: 85%;
+            max-height: 10.4em;
```
lgrahl: `10.4em` looks awkward in Firefox (half of the line is cut off) but `9.4em` works well enough for me in both Chrome and Firefox. Can you confirm that?

lgrahl:
Ping

src/sass/sections/_conversation.scss
joelfischerr: I thought the cut off text would indicate that the whole text is being quoted and not just the text that is being shown. But without being able to scroll through the text it does indeed look a bit awkward. `9.4em` works for me in Chrome, Firefox and Safari.

PR #769, opened 2019-03-05
Flow controlled data channels

### LINKED ISSUE #7: Allow sending large files
Currently file size is limited to 15 MiB because of an issue with WebRTC DataChannel buffering:

https://github.com/saltyrtc/saltyrtc-task-webrtc-js/issues/9

### LINKED ISSUE #594: Prevent implicit closing of WebRTC data channel
In Chrome (and it's derivatives), when the WebRTC data channel's buffering limit is being reached, the data channel implicitly closes when enqueuing more data. IIRC the limit was around 8 MiB.

This is one of the main issues responsible for #184.

https://github.com/saltyrtc/saltyrtc-task-webrtc-js/pull/21 fixes this but we also need to write this for the app side.

### LINKED ISSUE #741: Determine the maximum message size ourselves
We currently determine the (WebRTC) maximum message size as described by the negotiation mechanism in the SaltyRTC WebRTC task. However, this is superfluous and we can just use adapter.js for that purpose which will also be more accurate.

See https://github.com/saltyrtc/saltyrtc-meta/issues/130

lgrahl:
Resolves #594.
Closes #741 but is not effective until the app side has added support for the `v1` WebRTC task as well.
Resolves #7 (the app now also uses the `v1` WebRTC task with a flow control and out of memory handling, which is how we detect whether we can send large files).

To do:

- [x] Adjust code to be compatible with new chunking API
- [x] Adjust code to leverage the new SaltyRTC WebRTC Task API
  - Add a flow control to all data channels
- [x] Use released version of the SaltyRTC WebRTC Task in `package.json` (depends on https://github.com/saltyrtc/saltyrtc-task-webrtc-js/pull/28)

### DIFF
--- a/src/helpers/data_channel.ts
+++ b/src/helpers/data_channel.ts
@@ -1,0 +1,140 @@
+/**
+ * This file is part of Threema Web.
+ *
+ * Threema Web is free software: you can redistribute it and/or modify it
+ * under the terms of the GNU Affero General Public License as published by
+ * the Free Software Foundation, either version 3 of the License, or (at
+ * your option) any later version.
+ *
+ * This program is distributed in the hope that it will be useful, but
+ * WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the GNU Affero
+ * General Public License for more details.
+ *
+ * You should have received a copy of the GNU Affero General Public License
+ * along with Threema Web. If not, see <http://www.gnu.org/licenses/>.
+ */
+
+import {Logger} from 'ts-log';
+
+import {LogService} from '../services/log';
+
+/**
+ * A flow-controlled (sender side) data channel.
+ */
+export class FlowControlledDataChannel {
+    // Logging
+    private readonly log: Logger = null;
+
+    // Data channel
+    public readonly dc: RTCDataChannel;
+    public readonly highWaterMark: number;
+
+    // Flow control mechanism
+    private _ready: Future<void> = new Future();
+
+    /**
+     * Create a flow-controlled (sender side) data channel.
+     *
+     * @param dc The data channel to be flow-controlled.
+     * @param logService The logging service.
+     * @param logLevel The desired log level.
+     * @param lowWaterMark The low water mark unpauses the data channel
+     *   once the buffered amount of bytes becomes less or equal to it.
+     *   Defaults to 256 KiB.
+     * @param highWaterMark The high water mark pauses the data channel
+     *   once the buffered amount of bytes becomes greater or equal to it.
+     *   Defaults to 1 MiB.
+     */
+    public constructor(
+        dc: RTCDataChannel,
+        logService: LogService,
+        logLevel: threema.LogLevel,
+        lowWaterMark: number = 256 * 1024,
+        highWaterMark: number = 1024 * 1024,
+    ) {
+        this.log = logService.getLogger(
+            `FlowControlledDataChannel.${dc.id}`, 'color: #fff; background-color: #009688', logLevel);
+        this.dc = dc;
+        this.highWaterMark = highWaterMark;
+
+        // Allow writing
+        this._ready.resolve();
+
+        // Unpause once low water mark has been reached
+        this.dc.bufferedAmountLowThreshold = lowWaterMark;
+        this.dc.onbufferedamountlow = () => {
+            if (!this._ready.done) {
+                this.log.debug(`${this.dc.label} resumed (buffered=${this.dc.bufferedAmount})`);
+                this._ready.resolve();
+            }
+        };
+    }
+
+    /**
+     * A future whether the data channel is ready to be written on.
+     */
+    public get ready(): Future<void> {
+        return this._ready;
+    }
+
+    /**
+     * Write a message to the data channel's internal buffer for delivery to
+     * the remote side.
+     *
+     * Important: Before calling this, the `ready` Promise must be awaited.
+     *
+     * @param message The message to be sent.
+     * @throws Error in case the data channel is currently paused.
+     */
+    public write(message: Uint8Array): void {
+        // Throw if paused
+        if (!this._ready.done) {
+            throw new Error('Unable to write, data channel is paused!');
+        }
+
+        // Try sending
+        // Note: Technically we should be able to catch a TypeError in case the
+        //       underlying buffer is full. However, there are other reasons
+        //       that can result in a TypeError and no browser has implemented
+        //       this properly so far. Thus, we use a well-tested high water
+        //       mark instead and try to never fill the buffer completely.
+        this.dc.send(message);
+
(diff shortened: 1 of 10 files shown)


lgrahl:
Ready for review @dbrgn. :slightly_smiling_face:

src/helpers/data_channel.ts
```diff
@@ -0,0 +1,146 @@
...
+     * Write a message to the data channel's internal or application buffer for
+     * delivery to the remote side.
+     *
+     * @param message The message to be sent.
+     */
+    public write(message: Uint8Array) {
+        // Wait until ready, then write
+        // Note: This very simple technique allows for ordered message
+        //       queueing by using the event loop.
+        this.queue = this.queue.then(() => this.writeWhenReady(message));
```
lgrahl: :hugs:

src/helpers/data_channel.ts
```diff
@@ -0,0 +1,146 @@
...
+            throw new Error('Unable to write, data channel is paused!');
+        }
+
+        // Try sending
+        // Note: Technically we should be able to catch a TypeError in case the
+        //       underlying buffer is full. However, there are other reasons
+        //       that can result in a TypeError and no browser has implemented
+        //       this properly so far. Thus, we use a well-tested high water
+        //       mark instead and try to never fill the buffer completely.
+        this.dc.send(message);
```
dbrgn: We don't check the size of `message`, so if someone would send a 50 MiB message the DC would still close, right?

src/helpers/data_channel.ts
```diff
@@ -0,0 +1,146 @@
...
+     * @param $log The Angular logging service
+     * @param dc The data channel to be flow-controlled
+     * @param lowWaterMark The low water mark unpauses the data channel once
+     *   the buffered amount of bytes becomes less or equal to it.
+     * @param highWaterMark The high water mark pauses the data channel once
+     *   the buffered amount of bytes becomes greater or equal to it.
+     */
+    public constructor($log: ng.ILogService, dc: RTCDataChannel, lowWaterMark?: number, highWaterMark?: number) {
+        super($log, dc, lowWaterMark, highWaterMark);
+        this.queue = this.ready;
```
dbrgn: Again, when `this._ready` is replaced with a new future, won't the queue still reference the old future? Could this be a problem?

Probably I'm missing something, but as far as I can see, the future is never reset, but replaced. So once the promise in the root of the queue is resolved it would stay in resolved state forever.

src/helpers/data_channel.ts
dbrgn: Nice :slightly_smiling_face:

src/partials/welcome.ts
```diff
@@ -583,8 +583,7 @@ class WelcomeController {
             (error) => {
                 this.$log.error(this.logTag, 'Error state:', error);
                 // TODO: should probably show an error message instead
-                this.timeoutService.register(
-                    () => this.$state.reload(),
```
dbrgn: I'd try to avoid unrelated style changes (as long as tslint doesn't complain) to retain `git blame` functionality.

src/services/peerconnection.ts
```diff
@@ -174,12 +171,68 @@ export class PeerConnectionHelper {
...
+
+        // Bind events
+        dc.onopen = () => {
+            this.$log.info(this.logTag, `Signalling data channel open`);
+
+            // Rebind close event
+            dc.onclose = () => {
+                this.$log.info(this.logTag, `Signalling data channel closed`);
+                link.closed();
+            };
```
dbrgn: Just so I understand this: Why do we rebind onclose but not onerror or onmessage?

src/helpers/data_channel.ts
```diff
@@ -0,0 +1,146 @@
...
+        this.dc = dc;
+        this.highWaterMark = highWaterMark;
+
+        // Allow writing
+        this._ready.resolve();
+
+        // Unpause once low water mark has been reached
+        this.dc.bufferedAmountLowThreshold = lowWaterMark;
+        this.dc.onbufferedamountlow = () => {
+            if (!this._ready.done) {
```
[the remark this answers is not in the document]
lgrahl: Nah, it references `this` and fetches the `_ready` field on execution. The same would be the case for the getter, so it makes no difference.

src/helpers/data_channel.ts
lgrahl: The data channel has to do that check. If it doesn't, it's a bug (like in Chrome). If we want to work around this, then yeah, we would have to check it. However, it's never going to happen since we do chunking.

src/helpers/data_channel.ts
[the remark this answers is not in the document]
dbrgn: The getter is just syntactic sugar for a function call, so it does make a difference.

But probably the closure binding happens on the `this` instance, not on `this._ready`, I guess that's why it works.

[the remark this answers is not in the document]
lgrahl: > The getter is just syntactic sugar for a function call, so it does make a difference.

No. The function is also just a field.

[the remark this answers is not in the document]
dbrgn: Huh, you're right. It seems that closures always close over references to the lvalue (what do you call that in JS?). I though that they close over the rvalue.

```js
function main() {
  let a = Uint8Array.of(1, 2, 3);
  let b = 1;
  let arrow = () => {
    console.log(`arrow: a is ${a}, b is ${b}`);
  };
  function func() {
    console.log(`func: a is ${a}, b is ${b}`);
  };
  arrow();
  func();
  a = Uint8Array.of(4, 5, 6);
  b = 2;
  arrow();
  func();
}

main();
```

-> 

```
arrow: a is 1,2,3, b is 1
func: a is 1,2,3, b is 1
arrow: a is 4,5,6, b is 2
func: a is 4,5,6, b is 2
```

TIL.

src/helpers/data_channel.ts
lgrahl: Yes, in this case it references the `Future` that has been created on construction (see the constructor of `FlowControlledDataChannel`). That is the starting point of the queue. With each call to `write`, the `queue` is awaited by the `.writeWhenReady` `Promise` and immediately replaced by that promise.

Basically, `queue` is always the `Future` of the most recent `.write` call and resolves once it has written.

src/services/peerconnection.ts
lgrahl: The semantic of `close` changes (*should not happen* -> *valid*) after the channel opened but the semantic of `error` remains the same (*should not happen*).

dbrgn: Ah, of course :slightly_smiling_face:

lgrahl:
I might have found a way to resolve #7 but no promises. :slightly_smiling_face:

ovalseven8:
> I might have found a way to resolve #7 but no promises. slightly_smiling_face

Don't do that with me!! 😉😮🎉

PR #771, opened 2019-03-07
Process response/receivers after response/profile has been processed

### LINKED ISSUE #768: Setting a large profile picture will not allow a connection to Threema Web
## Expected Behavior
Setting a new profile picture in Threema Web should have no side effects except that the profile picture is updated.

## Current Behavior
When setting a new profile picture from a large jpeg the iOS app will not be able to reconnect on the next session. It will get stuck on 99%.

## Possible Solution
Check the size of the profile picture.

1. Set a new profile picture that is quite large such as this one in Threema Web: https://unsplash.com/photos/loGTCYD03n0
2. Refresh the page
3. Try to connect to Threema Web again
4. It will get stuck on 99% (at least on my connection, maybe it has something to do with connection speed)

## Your Environment
* Threema Web version: 2.1.7
* Threema App version: 4.1.2, 4.1.1
* Browser name and version: Firefox 65.0.1 and any other browser
* Computer operating system and version: macOS
* Smartphone operating system and version: iOS 12.1.4

lgrahl:
Resolves #768

### DIFF
--- a/src/services/webclient.ts
+++ b/src/services/webclient.ts
@@ -2172,11 +2172,16 @@ export class WebClientService {
             return future.reject('invalidResponse');
         }
 
-        // Store receivers
-        this.sortContacts(data.contact);
-        this.receivers.set(data);
-        this.registerInitializationStep(InitializationStep.Receivers);
-        future.resolve();
+        // Run delayed as it requires the profile to be initialised
+        this.runAfterInitializationSteps([
+            InitializationStep.Profile,
+        ], () => {
+            // Store receivers
+            this.sortContacts(data.contact);
+            this.receivers.set(data);
+            this.registerInitializationStep(InitializationStep.Receivers);
+            future.resolve();
+        });
     }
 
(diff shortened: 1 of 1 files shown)


src/services/webclient.ts
```diff
@@ -2172,11 +2172,16 @@ export class WebClientService {
             return future.reject('invalidResponse');
         }
 
-        // Store receivers
-        this.sortContacts(data.contact);
-        this.receivers.set(data);
-        this.registerInitializationStep(InitializationStep.Receivers);
-        future.resolve();
+        // Run delayed...
```
dbrgn: Maybe add a comment, *why* this depends on the profile :slightly_smiling_face:

lgrahl: "For technical reasons"? There's no real good rationale. It's just some assumption we make.

dbrgn: Something like this:

> // Run delayed, because the receiver processing code depends on the profile being initialized.

PR #772, opened 2019-03-07
Actually resize avatars

### LINKED ISSUE #770: Avatars should be resized to 512x512
## Expected Behavior

Avatars should be resized to 512x512.

## Current Behavior

At least when setting the profile, the avatar is not being resized and sent as-is to the app.

lgrahl:
Formerly, avatars have not been resized. Updated croppie type definitions so this doesn't happen again.

However, the race condition is still unfixed: When the user clicks on "Ok" before the timeout of 500ms fired, the avatar will not be updated. @dbrgn I have a patch ready if you see an easy way to communicate between the two controllers.

Resolves #770

### DIFF
--- a/src/directives/avatar_editor.ts
+++ b/src/directives/avatar_editor.ts
@@ -15,6 +15,9 @@
  * along with Threema Web. If not, see <http://www.gnu.org/licenses/>.
  */
 
+// tslint:disable:no-reference
+/// <reference path="../types/croppie.d.ts" />
+
 // tslint:disable:max-line-length
 
 import {bufferToUrl, logAdapter} from '../helpers';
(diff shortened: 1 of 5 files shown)


src/directives/avatar_editor.ts
```diff
@@ -41,47 +44,41 @@ export default [
...
+                                        circle: false,
+                                        format: 'png',
                                     });
-                            }, 500);
-                        },
-                    });
-
+                                    const fileReader = new FileReader();
+                                    fileReader.onload = function() {
+                                        scope.onChange(fileReader.result);
```
dbrgn: Just as a sidenote: `this` and `fileReader` are equivalent in the `onload` function, but I agree that using `fileReader` makes things more explicit.

lgrahl: They aren't the same with the refactoring any more. But yeah, let's avoid the brittleness of implicit `this` in general.

lgrahl:
I'll double-check that.

lgrahl:
`newImage(null)` is never being called.

PR #778, opened 2019-03-20
Markup parser: Ignore markup in URLs

dbrgn:
In the tokenizer, process URLs as text tokens. Use a different, smaller
set of boundary characters.

This avoids processing markup in URLs like
`https://example.com/file_-_1.jpg` (see test cases).

### DIFF
--- a/src/markup_parser.ts
+++ b/src/markup_parser.ts
@@ -53,8 +53,26 @@ function isMarkupToken(tokenType: TokenType) {
  * Return whether the specified character is a boundary character.
  * When `character` is undefined, the function will return true.
  */
-function isBoundary(character?: string) {
+function isBoundary(character?: string): boolean {
     return character === undefined || /[\s.,!?¡¿‽⸮;:&(){}\[\]⟨⟩‹›«»'"‘’“”*~\-_…⋯᠁]/.test(character);
+}
+
+/**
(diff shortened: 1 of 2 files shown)


src/markup_parser.ts
```diff
@@ -53,16 +53,35 @@ function isMarkupToken(tokenType: TokenType) {
...
+ */
+function isUrlBoundary(character?: string): boolean {
+    return character === undefined || !/[a-zA-Z0-9\-._~:/?#\[\]@!$&'()*+,;=%]/.test(character);
+}
+
+/**
+ * Return whether the specified string starts an URL.
+ */
+function isUrlStart(substring: string): boolean {
+    return substring.match(/^[a-zA-Z]+:\/\//) != null;
```
lgrahl: This requires that URLs always start with a scheme. Do we want that?

dbrgn: How else would you detect URLs? The problem is that we also want to support something like `Look at this *example.com*-crap!`. How would you determine whether the * in the middle is used for formatting or not?

I think it's quite improbable that someone sends an URL with unescaped characters in it that does not start with a scheme. Long URLs are usually copy-pasted, which includes the scheme.

lgrahl: Oh, right, we apply some post-processing to recognise domains, so this should be fine.

PR #784, opened 2019-04-02
Rename and clarify debugging flags in config

lgrahl:
:cat:


lgrahl:
Wake up @dbrgn. :grin:

dbrgn:
Can you rebase against master? It's likely that more usages of the renamed variables has been added in the meantime.

lgrahl:
Done.

PR #790, opened 2019-04-17
 Switch from Browserify to Webpack 

### LINKED ISSUE #539: Improve build toolchain
Currently we're using Browserify + Babel to generate an ES5 bundle.

Since browser support is better nowadays, we can generate ES2015 directly. For that, we should use [Babel Preset Env](https://babeljs.io/docs/en/babel-preset-env) with the browser versions we want to target.

When already looking at the build toolchain, we should also evaluate whether we might gain any advantages from moving to Webpack.

dbrgn:
Fixes #539


dbrgn:
Unfortunately this way we lose dynamic CSS injection support for now (unless we refactor the CSS build toolchain to use Webpack too), but we gain WebAssembly support.

Issue #753, opened 2019-02-17
New incoming message not shown


andyggg:
## Expected Behavior
New incoming messages should appear in the chat in Threema web.

## Current Behavior
Sometimes, new incoming messages don't appear in the chat. However, in most cases it works.

## Possible Solution
Work-around: disconnect and re-connect to Threema.

## Steps to Reproduce (for bugs)
1. Run Threema web in a tab In Firefox 65 (64 Bit)
2. Switch to another tab leaving the chat open
3. Surf the web in said other tab for 5(?) Minutes
4. receive a Threema message in the chat
5. Switch back to Threema-Tab
6. new message not shown in chat (sometimes! Most times it works)
7. Switching between chats doesn't help - chat does not refresh...

## Your Environment
* Threema Web version: 2.1.7
* Threema App version: 4.1.1 (2427)
* Browser name and version: Firefox 65 (64 Bit)
* Computer operating system and version: Windows 10
* Smartphone operating system and version: iOS 12.1.4

lgrahl:
Confirmed due to multiple users affected but we are still unable to reproduce it so far.

xan-a-a:
Just in case any of the Threema devs live in Berlin. I live like an hour away and experience this almost daily. I could call you as soon as it happens again so you could come over and have a look at it while I'm in this "useful state" of being connected yet not connected.

Or if there is anything I could do on my side to help you out, don't hesitate to tell me.

Version 72.0.3626.119 (Official Build) (64-bit)
macOS 10.14.3
iOS 12.1.2

dbrgn:
Are all affected users on iOS?

xan-a-a:
At least in my group chats, yes. I don't know how you keep the app active on Android, probably a service which is something that iOS doesn't offer. With Whatsapp Web I also get a lot of "Connection lost" messages especially on 2.4ghz Wifis - but at least it's telling me about it so I can hit the home button on my iPhone and connect normally again. It doesn't have to be a popup, WA Web just makes a yellow banner appear on top of the chat list telling you to stay connected. It then tries to reconnect after a couple of seconds - does Threema Web even try that? Or can it not do that because it's not noticing the connection loss? I've noticed the "`.`" behind "Threema" turning yellow or red sometimes but I haven't been able to make something out of it, seems rather random. 

[...]

PR #794, opened 2019-05-06
Chat message debugging

lgrahl:
Debug utilities to find the cause of #753.

Depends on #793

### DIFF
--- a/src/services/webclient.ts
+++ b/src/services/webclient.ts
@@ -23,7 +23,7 @@ import {StateService as UiStateService} from '@uirouter/angularjs';
 
 import * as msgpack from 'msgpack-lite';
 import {
-    arraysAreEqual, copyDeep, hasFeature, hasValue, hexToU8a,
+    arraysAreEqual, base64ToU8a, copyDeep, hasFeature, hasValue, hexToU8a,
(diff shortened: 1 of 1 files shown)


src/services/webclient.ts
```diff
@@ -4054,4 +4060,16 @@ export class WebClientService {
...
     }
+
+    /**
+     * Log chat message's metadata for debugging purposes.
+     */
+    private logChatMessages(type: string, subType: string, messages: threema.Message[]) {
+        for (const message of messages) {
+            const idHex = u8aToHex(base64ToU8a(message.id));
+            this.$log.debug('[MessageInspector]', `${type}/${subType}: direction=${message.isOutbox ? 'out' : 'in'}, ` +
+                `id=${idHex}. type=${message.type}, state=${message.state !== undefined ? message.state : '?'}, ` +
```
dbrgn: Should this `.` be a `,`?

```suggestion
                `id=${idHex}, type=${message.type}, state=${message.state !== undefined ? message.state : '?'}, ` +
```

lgrahl: Yes, good catch!

dbrgn: Or, as we say in Swiss German, "Tüpflischisserei" :smile:

PR #797, opened 2019-05-08
Chat message debugging enhanced

lgrahl:
Followup to #794


src/services/webclient.ts
```diff
@@ -2536,6 +2533,9 @@ export class WebClientService {
             this.$log.warn('Invalid messages response, unknown receiver type (' + type + ')');
             return future.reject('invalidResponse');
         }
+        if (this.config.MSG_DEBUGGING) {
+            this.logChatMessages(message.type, message.subType, type, id, 'new', data);
+        }
```
dbrgn: Maybe move this up a bit, so that invalid messages are logged too.

PR #805, opened 2019-05-15
Sort issue template sections more logically

rugk:
This has annoyed me some times already: If I want to report an issue, I e.g. certainly won't list "potential solutions" before even describing the steps to reproduce the issue…
Similarly, I e.g. often cannot describe a "current behaviour" when I did not even describe what happened before.

IMHO this new order is much more logical.

---

BTW GitHub also notifies one that this file uses the "legacy" version of issue templates (wrong location). You maybe want to change that, too: https://help.github.com/en/articles/about-issue-and-pull-request-templates


dbrgn:
Thanks. Opened an issue at https://github.com/threema-ch/threema-web/issues/806 for the new template structure.

PR #821, opened 2019-06-24
Fix collapsing of consecutive whitespace in compose area

dbrgn:
This was reverted previously because of bugs related to the old compose area. Now that we've merged the new one, the CSS change should work.

See #703.

### DIFF
--- a/README.md
+++ b/README.md
@@ -87,6 +87,7 @@ To run unit tests:
 
 To run UI tests:
 
+    npm run build  # Required for CSS to be rebuilt
     npm run test:ui <browser>
 
 For example:
--- a/package-lock.json
+++ b/package-lock.json
@@ -1232,9 +1232,9 @@
       }
     },
     "@threema/compose-area": {
-      "version": "0.3.2",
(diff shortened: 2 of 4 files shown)


dbrgn:
> Have you tested that it does work as expected?

Yes, of course. I'll take a look at the failing UI test.

dbrgn:
Seems to be a webdriver problem in Chrome. The test works if I replicate it manually. Somehow shift+enter doesn't seem to be processed correctly.

dbrgn:
Oh, now I know the actual reason. When `white-space: pre-wrap` is active, Chrome will not insert a `<br>` element when pressing shift+enter. Instead, it will insert multiple adjacent text nodes, which are rendered on separate lines if `white-space: pre-wrap` is active.

We could work around that by passing an argument to the compose area which indicates whether adjacent text nodes should be separated by newline or not. It means mixing display- and layout-information, but what else can we do if Chrome does these weird things :man_shrugging:

lgrahl:
If there's no way around it... but can we at least tie it to a concrete issue number in their bug tracker?

dbrgn:
Found a better solution :slightly_smiling_face: Apparently there are three text nodes, one of them contains nothing but a newline character. We can use that as newline indicator.

Issue #579, opened 2018-08-27
Report Tool


lgrahl:
We need a report tool to reduce the effort required to report logs to the support team and aggregate them with the corresponding app log.

ovalseven8:
Just tried it out. But I wonder if it should additionally be present in this menu:
![Bildschirmfoto vom 2019-08-02 13-10-28](https://user-images.githubusercontent.com/8258609/62408534-d7cbf500-b5ca-11e9-98ff-1f51f7428bb7.png)

lgrahl:
Good idea.

PR #826, opened 2019-06-27
Enhanced logging

lgrahl:
This completely replaces the Angular logging utilities with our own utilities. It is preliminary work for #579.

The log service creates a root logger which, by default, forwards all log records to two destinations:

- The console logger, and
- the in-memory logger intended to be used for reporting logs.

By default, the in-memory logger will keep the most recent log records, limiting to 1000 entries. It does not apply a log level filter.

The in-memory logger will serialise confidential log records concealed while the console logger will unveil confidential log records.

Also, colours, cleanup and more sophisticated ICE candidate censoring.


tsconfig.json
```diff
@@ -1,6 +1,6 @@
 {
     "compilerOptions": {
-        "target": "ES2015",
+        "target": "ES2017",
```
lgrahl: Needed for `String.prototype.padStart`.

src/helpers/logger.ts
```diff
@@ -0,0 +1,297 @@
...
+     *
+     * @param space Amount of white spaces used for nested block indentation.
+     */
+    public serialize(space: number = 2): string {
+        const records = this.records.map(([date, type, message, ...args]: LogRecord) => {
+            // Strip message formatting
+            if (message.constructor === String) {
+                let stripped = false;
+
+                // Strip style formatting string if any
```
dbrgn: Maybe re-word to

```suggestion
                // Strip first style formatting placeholder if any
```

src/config.ts
```diff
@@ -38,13 +37,22 @@ export default {
...
+    DEBUG_TIMER: false,
+    // Toggles logging all chunks and messages exchanged by or associated with
+    // the app remote protocol. May also enable additional debug facilities of
+    // the app remote protocol.
+    // Note: Affects performance and contains sensitive information.
+    DEBUG_ARP: false,
+    // Toggles URL logging to visualise MsgPack messages for all incoming and
+    // outgoing protocol messages.
+    // Note: Affects performance and contains sensitive information.
+    DEBUG_MSGPACK: false,
```
dbrgn: Should we split this up into two objects?

```ts
LOG_LEVEL: {
  CONSOLE: 'info',
  COMPOSE_AREA: 'warn',
  SALTYRTC: 'warn',
},
DEBUG: {
  TIMER: false,
  ARP: false,
  MSGPACK: false,
}
```

I think that would make it much easier to read when adding more entries. On the other hand patching the config is much harder because the values are namespaced on another line of code...

src/helpers/confidential.ts
```diff
@@ -0,0 +1,169 @@
...
+    // Filter binary data
+    if (value instanceof ArrayBuffer) {
+        return `[ArrayBuffer: length=${value.byteLength}]`;
+    }
+    if (value instanceof Uint8Array) {
+        return `[Uint8Array: length=${value.byteLength}, offset=${value.byteOffset}]`;
+    }
+    if (value instanceof Blob) {
+        return `[Blob: length=${value.size}, type=${value.type}]`;
+    }
```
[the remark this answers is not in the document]
lgrahl: Yup but since we have no use case for them and probably never will have, I went the easy way.

src/helpers/confidential.ts
```diff
@@ -0,0 +1,169 @@
...
+
+    // Apply filter to confidential data
+    if (value instanceof BaseConfidential) {
+        return value.censored();
+    }
+
+    // Filter boolean and number
+    if (value.constructor === Boolean || value.constructor === Number) {
+        return `[${value.constructor.name}]`;
+    }
```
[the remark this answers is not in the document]
lgrahl: Good point.

src/helpers/logger.ts
```diff
@@ -0,0 +1,297 @@
...
+    public readonly info: LogFunction = this.noop;
+    public readonly warn: LogFunction = this.noop;
+    public readonly error: LogFunction = this.noop;
+
+    constructor(logger: Logger, level: LogLevel) {
+        this.logger = logger;
+        this.level = level;
+
+        // Bind corresponding method to log level type, if enabled
+        // noinspection FallThroughInSwitchStatementJS
```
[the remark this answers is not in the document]
lgrahl: The fall-through is intentional here and it would actually break if there were a `break`.

src/helpers/logger.ts
```diff
@@ -0,0 +1,297 @@
...
+     * Drops the oldest log record if the log record limit would be exceeded.
+     *
+     * @param type The log record level type.
+     * @param message The log message.
+     * @param args Further arguments of the log record
+     */
+    private append(type: LogType, message?: any, ...args: any[]): void {
+        // Remove oldest record if needed
+        if (this.limit > 0 && this.records.length >= this.limit) {
+            this.records.shift();
```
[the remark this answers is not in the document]
lgrahl: Yeah, let's not do premature optimisation because fixing this would be slightly complicated. :slightly_smiling_face:

src/helpers/logger.ts
```diff
@@ -0,0 +1,297 @@
...
+        }
+
+        // Add newest record
+        this.records.push([new Date(), type, message, ...args]);
```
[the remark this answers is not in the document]
lgrahl: Well, I agree and I don't agree. Object keys are not guaranteed to be ordered, so readability is not guaranteed. In the end, we should parse the JSON and display it in a tiny debugging tool anyway.

src/helpers/logger.ts
```diff
@@ -0,0 +1,297 @@
...
+     *
+     * While serialising, a recursive filter will be applied:
+     *
+     * - the types `null`, `string`, `number` and `boolean` will be returned
+     *   unmodified,
+     * - an object implementing the `Confidential` interface will be returned
+     *   sanitised,
+     * - an `Error` instance will be left as is,
+     * - the binary types `Uint8Array` and `Blob` will only return meta
+     *   information about the content, and
```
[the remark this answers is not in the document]
lgrahl: Yup, keep it simple.

src/config.ts
lgrahl: My idealistic mind agrees but my practical mind thinks that the config is still well readable atm.

src/helpers/logger.ts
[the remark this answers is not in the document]
dbrgn: Why? The level cannot be `debug` and `info` at the same time.

src/helpers/logger.ts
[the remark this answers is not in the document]
dbrgn: I was talking about code, not about the log data in the console. The line `console.log(record.type)` is much easier to read than `console.log(record[1])`.

But I leave it to do whether to keep the array based approach or not. It's still typed after all.

src/config.ts
dbrgn: Let's keep it, because of the config-patching issue :slightly_smiling_face:

src/helpers/logger.ts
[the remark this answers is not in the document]
lgrahl: This is not how it works. It binds every function equal or above `level`:

- `debug` binds `debug`, `trace`, `info`, `warn` and `error`,
- `warn` binds `warn` and `error`.

src/helpers/logger.ts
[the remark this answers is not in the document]
lgrahl: Yep. And `args` is easier to use with an array, so I think we should stick to it.

src/helpers/logger.ts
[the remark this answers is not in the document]
dbrgn: Holy crap, I just realized that I didn't know that fallthrough works this way. I thought that the condition is still evaluated if there is code in between two conditions... This is horrible!

Anyways, I guess it's fine if you assume that people reading this know JS (unlike me apparently). Although in that case I don't see any use in catching the `default` arm.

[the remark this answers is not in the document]
lgrahl: > Holy crap, I just realized that I didn't know that fallthrough works this way. I thought that the condition is still evaluated if there is code in between two conditions... This is horrible!

Your openness is admirable and you've learned something. :tada:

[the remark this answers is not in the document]
lgrahl: I agree that there aren't many valid use cases of fall-through, similar to `goto`, but this is one of the valid ones IMO.

src/helpers.ts
```diff
@@ -371,18 +371,61 @@ export function isActionTrigger(ev: KeyboardEvent): boolean {
...
+ * This handles the following types:
+ *
+ * - copies `undefined` and `null`,
+ * - copies `Boolean`, `Number` and `String`,
+ * - copies `object` recursively,
+ * - copies `Array` recursively,
+ * - copies `ArrayBuffer`,
+ * - copies `Uint8Array`,
+ *
+ * Everything else will be **referenced**.
```
dbrgn: Isn't that totally against the intent of a deep copy?

src/helpers.ts
```diff
@@ -282,28 +284,26 @@ export function msgpackVisualizer(array: Uint8Array): string {
...
         // tslint:disable:no-bitwise
         return (contactReceiver.featureMask & feature) !== 0;
         // tslint:enable:no-bitwise
     }
-    $log.warn(logTag, 'Cannot check featureMask of a undefined contactReceiver');
+    log.warn('Cannot check featureMask of a undefined contact receiver');
     return false;
 }
 
+const FALLBACK_MIME_TYPE = 'image/jpeg';
```
dbrgn: If this constant is moved out of the function, the context isn't clear anymore. Other helper functions won't utilize it. Either add a comment explaining the constant, or (better in my opinion) move the constant definition into the function body itself.

src/helpers/confidential.ts
```diff
@@ -167,3 +169,56 @@ export class ConfidentialWireMessage extends BaseConfidential<threema.WireMessag
...
+        const ipv4 = ip.split('.');
+        if (ipv4.length > 1) {
+            return `${ipv4.slice(0, 2).join('.')}.*.*`;
+        }
+
+        // Handle IPv6 address (catch-all)
+        const ipv6 = ip.split(':');
+        const head = ipv6.shift();
+        return `${head}:${ipv6.map((item) => item.length > 0 ? '*' : item).join(':')}`;
+    }
```
dbrgn: This would be an ideal candidate (ha-ha) for a unit test.

src/helpers.ts
lgrahl: Considering that it was not copied before either but just represented as an empty object, this already is an improvement. I'm happy to bikeshed on the name. Suggestions?

lgrahl: You have to see it in the context we're using it. That is currently solely for logging binary data that could be modified later. This is not meant to be a general purpose deep copy function that covers all edge cases. The catch-all currently handles copying of all non-special types. Throwing an exception instead (which is an option) would require checking more types.

lgrahl:
Tests are mostly done. Only the `LogService` needs some and perhaps the `copyDeep` helper.

tests/ts/confidential_helpers.ts
```diff
@@ -0,0 +1,303 @@
...
+                subType: 'dessert',
+                args: 'arrrrrrgggsss',
+                data: {
+                    preference: ['ice cream', 'chocolate'],
+                    priority: Number.POSITIVE_INFINITY,
+                },
+            };
+            const confidential = new ConfidentialWireMessage(message);
+            expect(confidential.censored()).toEqual({
+                type: 'request/food',
```
lgrahl: Ah, this is actually not a valid ARP type... but who cares.

src/helpers.ts
dbrgn: That's fine, but then you need to change the name. I don't have a good idea though, `copySemideep`? :stuck_out_tongue: 

The fallback could also be serialize-deserialize, as we did for the entire object previously.

tests/ts/logger_helpers.ts
```diff
@@ -0,0 +1,511 @@
...
+ * Format a log record. This splices a tag into the expected place.
+ */
+function formatLogRecord(record: LogRecord, tag?: string[]): LogRecord {
+    if (tag === undefined) {
+        return record;
+    }
+
+    // Splice tag into message
+    const [type, ...args] = record;
+    return [type, ...tag.concat(args)];
```
dbrgn: It took a few seconds for me to parse this... This is something where `record.tag = tag` would be *much* easier to read :slightly_smiling_face:

src/helpers.ts
lgrahl: > The fallback could also be serialize-deserialize, as we did for the entire object previously.

That adds zero value since it just returns an empty object.

tests/ts/logger_helpers.ts
```diff
@@ -0,0 +1,511 @@
...
+            const root = new UnveilLogger(logger);
+
+            // Ensure a confidential record is being unveiled
+            root.debug(new TestConfidential());
+            expect(logger.records).toContain(['debug', 'uncensored']);
+        });
+    });
+
+    describe('ConsoleLogger', () => {
+        let backup: Logger = {} as Logger;
```
dbrgn: Why do you cast this to `Logger`? Isn't this a plain `object` containing loggers as values?

tests/ts/logger_helpers.ts
```diff
@@ -0,0 +1,511 @@
...
+
+            // Store each log level type method of console that we will override
+            for (const type of LOG_TYPES) {
+                backup[type] = console[type];
+            }
+
+            // Overwrite each log level type method of console
+            for (const type of LOG_TYPES) {
+                console[type] = logger[type];
+            }
```
dbrgn: Alternatively you can also use spies: https://www.htmlgoodies.com/html5/javascript/spy-on-javascript-methods-using-the-jasmine-testing-framework.html

Something like this (untested):

```
spyOn(console, "debug");
...
expect(console.debug).toHaveBeenCalled();
```

No need to change the current test code, but might simplify some tasks in the future.

src/helpers.ts
dbrgn: Only for non-serializable types (like ArrayBuffer), of which we explicitly handle a few. But I'd prefer the renaming too.

`copyDeeper` could also be a potential name.

lgrahl: > Only for non-serializable types

There aren't any further types. JSON is, by definition, limited to JSON.

How about `copyDeepish`?

tests/ts/logger_helpers.ts
lgrahl: Good idea!

tests/ts/logger_helpers.ts
lgrahl: This would require squeezing in-band meta-information through the interface defined by `ts-log` which I'm not willing to do.

tests/ts/logger_helpers.ts
lgrahl: This holds the `Console` log level methods until they are being restored. I could leave it as an object but since it actually does implement the interface, it doesn't hurt.

src/helpers.ts
dbrgn: `copyMost`?

tests/ts/logger_helpers.ts
dbrgn: Ahh, I didn't relalize that it *does* implement the interface. In that case it makes sense of course :slightly_smiling_face:

src/helpers.ts
lgrahl: `copyDeepExceptWhereItReferences`.

PR #830, opened 2019-07-03
Add transifex-client config

dbrgn:
This way we can pull down translations (with at least 90% covered) using `tx pull --all`.


dbrgn:
@lgrahl updated with link to transifex-client.

PR #835, opened 2019-07-04
Fix sending files and pictures to distribution lists

### LINKED ISSUE #394: Sending pictures to distribution lists won't work
## Expected Behavior
When sending pictures to a distribution list created in the threema app, the pictures should be send to all members of this list.

## Current Behavior
Sending pictures through the web client to distribution lists is not working. 

1. Create a distribution list on threema
2. Log in to the web client
3. Send a picture to the newly created distribution list through the web client

## Your Environment
* Threema Web version: 1.7.0
* Threema for Android version: 3.3.2000391
* Browser name and version: Chromium 62.0.3202.94
* Operating system and version: Manjaro Linux 17.0.6

dbrgn:
Fixes #394.

### DIFF
--- a/src/services/webclient.ts
+++ b/src/services/webclient.ts
@@ -1752,18 +1752,29 @@ export class WebClientService {
 
                         // check receiver
                         switch (receiver.type) {
-                            case 'distributionList':
-                                return reject(this.$translate.instant(
-                                    invalidFeatureMessage, {receiverName: receiver.displayName},
-                                ));
                             case 'group':
+                            case 'distributionList':
                                 const unsupportedMembers = [];
-                                const group = this.groups.get(receiver.id);
-
-                                if (group === undefined) {
-                                    return reject();
+                                let members: string[];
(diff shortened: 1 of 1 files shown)


src/services/webclient.ts
```diff
@@ -1752,18 +1752,29 @@ export class WebClientService {
...
+                                        const distributionList = this.distributionLists.get(receiver.id);
+                                        if (distributionList === undefined) {
+                                            this.log.error(`Distribution list ${receiver.id} not found`);
+                                            return reject(this.$translate.instant('error.ERROR_OCCURRED'));
+                                        }
+                                        members = distributionList.members;
+                                        break;
                                 }
-                                group.members.forEach((identity: string) => {
+                                members.forEach((identity: string) => {
```
lgrahl: Any reason to not make this is a normal `for`?

dbrgn: I don't think so. I'll change it.

PR #837, opened 2019-07-04
Allow adding all contacts to distribution lists

### LINKED ISSUE #836: Distribution Lists: Cannot add members without GROUP_CHAT capability
## Bug Description

I cannot add members without the group chat feature flag to distribution lists, even though that should be allowed.

- Adding ECHOECHO to groups should not work
- Adding ECHOECHO to distribution lists should work

dbrgn:
Fixes #836.

### DIFF
--- a/src/controller_model/distributionList.ts
+++ b/src/controller_model/distributionList.ts
@@ -30,7 +30,7 @@ export class DistributionListControllerModel implements threema.ControllerModel<
     public name: string;
     public subject: string;
     public isLoading = false;
-    public readonly requiredMemberFeatureMask = null; // No constraints
+    public readonly requiredMemberFeatureMask: threema.ContactReceiverFeature = null; // No capability constraints
 
     private addContactPlaceholder: string;
     private distributionList: threema.DistributionListReceiver | null;
--- a/src/controller_model/distributionList.ts
+++ b/src/controller_model/distributionList.ts
@@ -22,7 +22,9 @@ import {WebClientService} from '../services/webclient';
 // Type aliases
 import ControllerModelMode = threema.ControllerModelMode;
 
-export class DistributionListControllerModel implements threema.ControllerModel<threema.DistributionListReceiver> {
+export class DistributionListControllerModel
+        implements threema.ControllerModel<threema.DistributionListReceiver>,
+                   threema.ControllerModelWithMembers {
     private $translate: ng.translate.ITranslateService;
     private $mdDialog: ng.material.IDialogService;
     private readonly log: Logger;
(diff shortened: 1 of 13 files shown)


src/controller_model/distributionList.ts
```diff
@@ -30,6 +30,7 @@ export class DistributionListControllerModel implements threema.ControllerModel<
     public name: string;
     public subject: string;
     public isLoading = false;
+    public readonly requiredMemberFeatureMask = null; // No constraints
```
lgrahl: Do you mean `any` by that? But it clearly isn't `any`.

src/controller_model/group.ts
```diff
@@ -31,6 +31,7 @@ export class GroupControllerModel implements threema.ControllerModel<threema.Gro
     public name: string;
     public access: threema.GroupReceiverAccess;
     public subject: string;
+    public readonly requiredMemberFeatureMask = threema.ContactReceiverFeature.GROUP_CHAT;
```
lgrahl: Type missing.

src/controller_model/distributionList.ts
dbrgn: I can't follow. What do you mean?

dbrgn:
@lgrahl type added to an interface.

src/helpers.ts
```diff
@@ -290,6 +290,9 @@ export function hasFeature(contactReceiver: threema.ContactReceiver,
             log.warn(`Contact receiver with id ${contactReceiver.id} has featureMask 0`);
             return false;
         }
+        if (feature === threema.ContactReceiverFeature.NONE) {
+            return true;
+        }
```
dbrgn: :confused:

dbrgn:
@lgrahl refactoring done.

src/helpers.ts
lgrahl: Slightly awkward but not any more awkward than the `undefined` check above. :smirk_cat:

src/controller_model/distributionList.ts
```diff
@@ -22,18 +22,28 @@ import {WebClientService} from '../services/webclient';
...
     private $mdDialog: ng.material.IDialogService;
+
+    // Custom services
     private readonly log: Logger;
-    public members: string[];
-    public name: string;
+    private readonly webClientService: WebClientService;
+
+    // Fields required by interface
+    public readonly receiverType = 'contact';
```
dbrgn: Oh nice. How did this not break something when testing it?!

One of the nice side effects of trying to add more structure, it increases the surface for bugs.

dbrgn: No failure because unused (messenger uses this for mentions, which are only allowed in groups).

PR #838, opened 2019-07-04
Improved troubleshooting

lgrahl:
We now check that the peer-to-peer connection can be established and that data channels are working as expected.

### DIFF
--- a/troubleshoot/troubleshoot.js
+++ b/troubleshoot/troubleshoot.js
@@ -1,4 +1,158 @@
-var app = angular.module('troubleshoot', ['ngSanitize']);
+/**
+ * Create peer connection and bind events to be logged.
+ * @param role The role (offerer or answerer).
+ * @returns {RTCPeerConnection}
+ */
+function createPeerConnection(role) {
+    // Detect safari
+    const uagent = window.navigator.userAgent.toLowerCase();
+    const isSafari  = /safari/.test(uagent) && /applewebkit/.test(uagent) && !/chrome/.test(uagent);
+
+    // Determine ICE servers
+    let iceServers;
+    if (isSafari) {
+        iceServers = [
+            'turn:turn.threema.ch:443?transport=udp',
+            'turn:turn.threema.ch:443?transport=tcp',
+            'turns:turn.threema.ch:443',
+        ];
+    } else {
+        iceServers = [
+            'turn:ds-turn.threema.ch:443?transport=udp',
+            'turn:ds-turn.threema.ch:443?transport=tcp',
+            'turns:ds-turn.threema.ch:443',
+        ];
+    }
+    console.debug('Using ICE servers: ' + iceServers);
+    const configuration = {iceServers: [{
+        urls: iceServers,
+        username: 'threema-angular-test',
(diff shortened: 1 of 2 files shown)


troubleshoot/troubleshoot.js
```diff
@@ -225,84 +330,258 @@ app.controller('ChecksController', function($scope, $timeout) {
...
+                    } catch (error) {
+                        this.resultDc.logs.push(
+                            `${role}: Channel '${dc.label}' was unable to send (${error.toString()})`);
+                        reject();
+                    }
                 });
+            });
+            dc.addEventListener('close', () => {
+                $scope.$apply(() => {
+                    if (this.resultDc.state  === 'unknown') {
```
lgrahl:
```suggestion
                    if (this.resultDc.state  === 'loading') {
```

troubleshoot/troubleshoot.js
```diff
@@ -1,4 +1,158 @@
...
+
+    // Wait until all is done
+    await Promise.all([
+        offererConnected,
+        answererConnected,
+        signalingDone,
+    ]);
+}
+
+// Here beginneth the Angular stuff
```
dbrgn: Heyday! Thou dost wield an enchanting tongue!

troubleshoot/troubleshoot.js
```diff
@@ -183,9 +290,7 @@ app.controller('ChecksController', function($scope, $timeout) {
             console.log('Message bytes:', bytes);
 
             // Validate length
-            let valid;
             if (bytes.length < 81) {
-                valid = false;
```
dbrgn: :smile: I wonder where that comes from.

troubleshoot/troubleshoot.js
```diff
@@ -225,84 +330,258 @@ app.controller('ChecksController', function($scope, $timeout) {
...
+                    .then(() => new Promise((resolve) => setTimeout(resolve, 1000)))
+                    .then(() => new Promise((resolve, reject) => {
+                        const dc = createDataChannel(offerer, 'Offerer', 'therme');
+                        canUseDataChannel('Offerer', dc, resolve, reject);
+                    })),
+                new Promise((resolve, reject) => {
+                    answerer.addEventListener('datachannel', (event) => {
+                        $scope.$apply(() => {
+                            const dc = event.channel;
+                            if (dc.label !== 'therme') {
```
dbrgn: This magic value could be moved to a constant.

dbrgn:
Additional comment: The fact that more debug logs are printed now is nice, but it means that most people won't be able to create a screenshot anymore (whereas before that was the way most people transmitted the results).

I wonder if we should have a copy-pasteable text summary somewhere.

Alternatively, maybe we could make the layout more compact, e.g. with two columns.

lgrahl:
Nitpickery... nice job description! :grin:

lgrahl:
> Additional comment: The fact that more debug logs are printed now is nice, but it means that most people won't be able to create a screenshot anymore (whereas before that was the way most people transmitted the results).
> 
> I wonder if we should have a copy-pasteable text summary somewhere.
> 
> Alternatively, maybe we could make the layout more compact, e.g. with two columns.

Kind of agree but then I don't think it's too much to ask to let people make two screenshots. :slightly_smiling_face:

PR #839, opened 2019-07-04
Report Tool / Troubleshooting

### LINKED ISSUE #579: Report Tool

lgrahl:
Adds a report tool for troubleshooting. This allows sending a log to `*SUPPORT` or copy it to the clipboard.

Depends on #858
Resolves #579 

To do:

* [x] Copy the report log into the clipboard
* [x] Send the report log to `*SUPPORT`
  - [x] Re-establish connection (if needed)
  - [x] Add contact `*SUPPORT`
  - [x] Send as file to `*SUPPORT`
  - [x] Require a problem description
  - [x] If anything fails, fall back to web form
* [x] Resolve bugs related to sending
* [x] Add metadata
  - [x] Web Client version
  - [x] Browser and version
  - [x] Configuration
  - [x] ~~Settings~~
* [x] Add log viewer tool

### DIFF
--- a/src/helpers/clipboard.ts
+++ b/src/helpers/clipboard.ts
@@ -1,0 +1,65 @@
+/**
+ * This file is part of Threema Web.
+ *
+ * Threema Web is free software: you can redistribute it and/or modify it
+ * under the terms of the GNU Affero General Public License as published by
+ * the Free Software Foundation, either version 3 of the License, or (at
+ * your option) any later version.
+ *
+ * This program is distributed in the hope that it will be useful, but
+ * WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the GNU Affero
+ * General Public License for more details.
+ *
+ * You should have received a copy of the GNU Affero General Public License
+ * along with Threema Web. If not, see <http://www.gnu.org/licenses/>.
+ */
+
+/**
+ * Copy a string into the clipboard.
+ *
+ * @param text The string to be copied.
+ * @param useSelectionRange whether a selection range should be used. Required
+ *   for Safari.
+ *
+ * Throws an error in case it was unsuccessful.
+ */
+export function copyString(text: string, useSelectionRange: boolean = false): void {
+    const activeElement = document.activeElement as HTMLElement | null;
+
+    // Create temporary (and hidden) textarea element
+    const textArea = document.createElement('textarea');
+    textArea.value = text;
+    textArea.style.visibility = 'hidden';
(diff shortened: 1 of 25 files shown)


lgrahl:
This is mostly done now and ready for review. I'm only thinking about adding additional metadata.

lgrahl:
Works well now. Only metadata left.

lgrahl:
Done. Ready for review @dbrgn.

troubleshoot/log.html
```diff
@@ -0,0 +1,196 @@
...
+
+        <h2>Log</h2>
+        <table id="log"></table>
+    </div>
+</div>
+
+<footer>&copy; 2017&ndash;2019 Threema GmbH</footer>
+
+<!-- JS -->
+<script src="log.js?v=[[VERSION]]"></script>
```
dbrgn: You need to add this file path to `dist/package.sh` (line 103 I think), in order for the [[VERSION]] string to be replaced.

dbrgn:
Maybe remove the TURN credentials from the generated logfile.

src/helpers/clipboard.ts
```diff
@@ -0,0 +1,57 @@
...
+        textArea.readOnly = false;
+        const range = document.createRange();
+        const selection = self.getSelection();
+        selection.removeAllRanges();
+        selection.addRange(range);
+        textArea.setSelectionRange(0, 999999);
+    } else {
+        textArea.focus();
+        textArea.select();
+    }
```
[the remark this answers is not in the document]
lgrahl: I went with restoring the focus since that's cleaner.

src/sass/base/_clipboard.scss
```diff
@@ -0,0 +1,6 @@
+// Hide temporary textareas used for copying into the clipboard
+body {
+    > textarea {
+        visibility: hidden;
+    }
+}
```
dbrgn: Well, this works, but I meant setting inline CSS with JavaScript :slightly_smiling_face: 

```javascript
element.style.display = 'none';
```

Adding a global CSS rule with its own CSS file is a bit overkill in my opinion and - while a top-level textarea is highly unlikely - might have side effects.

lgrahl: Sigh...

dbrgn: Yes, but less files in less locations = higher cohesion and easier maintenance :slightly_smiling_face:

lgrahl: It depends on our global CSS anyway unless you want to prevent rendering it... which could (and probably will) be optimised away and ignored by the browser.

PR #841, opened 2019-07-05
Add link checker to pipeline

ovalseven8:
Check URLs

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -44,6 +44,14 @@ jobs:
     environment:
       BUILDTARGET: build:uitests
       TESTTARGET: test:ui chrome
+  link-check:
+    docker:
+      - image: docker:stable
+    steps:
+      - run: apk update && apk add git grep
+      - checkout
+      - setup_remote_docker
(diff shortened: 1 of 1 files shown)


ovalseven8:
Hey, no idea if you like this idea. :)

Using `grep` and `sed` you can exactly specify which links to check. At the moment, all links that are not in the `tests` directory or in a hidden directory are checked. But you could limit that if you want.

For example, sometimes you put links into the source code for description and this script would check if it's still a valid link. It is using a Python script that does HEAD requests for all the URLs. In case the HEAD requests returns a `404` return code it is checked again using a GET request because some servers are configured to only allow GET but not HEAD. Open for ideas though.

Also see https://github.com/ovalseven8/simple-link-checker

dbrgn:
Hi, thanks for the suggestion.

I'm not sure if it's worth doing, since we have almost no user-facing URLs in the application (as opposed to – for example – a website like MDN that contains thousands of URLs). It might also add more friction to contributions, since CI failure now depends on external factors (if a file contains an URL to a server that has a hiccup, the tests will fail even though the PR is fine). We'd also have to filter out Transifex URLs, since the profiles are only visible with a login. However, I'm not generally opposed to it.

@lgrahl what's your opinion?

ovalseven8:
One could specify to only check URLs in the `src` directory for example.

lgrahl:
I think this is useful. However, I have a suggestion: Instead of relying on a third-party Docker image, and since we have a lot of npm tools anyway, can we use a small Node utility for that?

PR #854, opened 2019-07-15
Close peer connection if the connection state is 'disconnected' for 15s

### LINKED ISSUE #844: Connection lost without reconnection attempt
In the current beta:

image

img

That was 10 minutes ago. @lgrahl any idea why the push wasn't triggered?

lgrahl:
Previously, we relied on the ICE connection state going into 'failed' (or 'closed') eventually. However, with continuous gathering, there is no guarantee this ever happens. Furthermore, this may take an arbitrary amount of time. Recently, libwebrtc has turned off dispatching of the 'failed' state completely, see: https://bugs.chromium.org/p/chromium/issues/detail?id=982793

Also, a bit of cleanup (sorry).

Resolves #844

### DIFF
--- a/src/services/peerconnection.ts
+++ b/src/services/peerconnection.ts
@@ -107,7 +107,8 @@ export class PeerConnectionHelper {
             }
         };
         this.pc.onicecandidateerror = (e: RTCPeerConnectionIceErrorEvent) => {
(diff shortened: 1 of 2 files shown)


dbrgn:
Looks good, thanks. I'll test it for a few hours before merging.

PR #856, opened 2019-07-22
Recurrent wakeup for unsent messages

lgrahl:
Following the changes of https://github.com/threema-ch/app-remote-protocol/pull/5.

Related (but does not resolve) #753.


src/threema/container.ts
```diff
@@ -711,24 +711,25 @@ class Messages implements threema.Container.Messages {
         return false;
     }
 
-    public bindTemporaryToMessageId(receiver: threema.BaseReceiver, temporaryId: string, messageId: string): boolean {
+    public bindTemporaryToMessageId(
+        receiver: threema.BaseReceiver, temporaryId: string, messageId: string): threema.Message | null {
```
dbrgn: Just a note on formatting: If line breaks are inevitable, I prefer a style similar to the one used by rustfmt:

```typescript
public bindTemporaryToMessageId(
    receiver: threema.BaseReceiver,
    temporaryId: string,
    messageId: string,
): threema.Message | null {
```

Main advantage is that it keeps git history clean because a change to a single argument doesn't modify the entire line.

But feel free to ignore this comment, I don't have a strong opinion on this.

src/threema/container.ts
```diff
@@ -711,24 +711,25 @@ class Messages implements threema.Container.Messages {
         return false;
     }
 
-    public bindTemporaryToMessageId(receiver: threema.BaseReceiver, temporaryId: string, messageId: string): boolean {
+    public bindTemporaryToMessageId(
```
dbrgn: Can you add a docstring? I had to read through the function body to remember what the function does.

src/services/webclient.ts
```diff
@@ -2590,12 +2590,20 @@ export class WebClientService {
...
+        const message = this.messages.bindTemporaryToMessageId(
             receiver,
-            message.ack.id,
+            wireMessage.ack.id,
             messageId,
         );
         future.resolve(messageId);
+
+        // Add a special future that resolves once the message has been
+        // identified as sent.
```
dbrgn: Maybe expand this a bit:

```suggestion
        // identified as sent. As long as an unacknowledged wire message future exists, the app will be woken up again on connection loss.
```

(With proper line breaks of course.)

src/threema/container.ts
lgrahl: I do agree though that it should be done consistently. I struggle with that in my own projects and have only found a viable solution for it for C projects where the code is being automatically reformatted with `clang-format`.

PR #872, opened 2019-08-06
Fix `sortKey` for temporary messages

### LINKED ISSUE #871: Messages not properly marked as read
## Bug Description

Occasionally messages received from a contact are not marked as read. The wrong message ID is sent (too low).

This only happens in Beta 4, not in Beta 3.

Code that marks messages as read can be found in `messenger.ts`.

lgrahl:
Resolves #871

### DIFF
--- a/src/threema/container.ts
+++ b/src/threema/container.ts
@@ -567,16 +567,31 @@ class Messages implements threema.Container.Messages {
      * Messages must be sorted ascending by date.
      */
     public addNewer(receiver: threema.BaseReceiver, messages: threema.Message[]): void {
+        // Ignore empty list
         if (messages.length === 0) {
-            // do nothing
             return;
         }
         const receiverMessages = this.getReceiverMessages(receiver);
-        // if the list is empty, add the current message as ref
+
+        // Set oldest message if none previously available
         if (receiverMessages.list.length === 0) {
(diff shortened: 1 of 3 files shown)


src/threema/container.ts
```diff
@@ -567,16 +567,31 @@ class Messages implements threema.Container.Messages {
...
+
+        // Append new messages
+        for (const message of messages) {
+            // Determine sort key (if not set)
+            // Note: This resolves an ugly interface violation of temporary
+            //       messages.
+            if (message.sortKey === undefined) {
+                message.sortKey = receiverMessages.list.length > 0
+                    ? receiverMessages.list[receiverMessages.list.length - 1].sortKey
+                    : 0;
```
dbrgn: Note that in theory the sort key of an actual message could be negative, since the field is just an integer.

Maybe `Number.MIN_SAFE_INTEGER`?

Alternatively update the ARP and state that values may not be negative.

lgrahl: The latter makes more sense to me.

lgrahl: https://github.com/threema-ch/app-remote-protocol/pull/6

lgrahl:
@dbrgn Good to merge then?

PR #875, opened 2019-08-06
Allow to opt-in sensitive data when using the report tool

lgrahl:
Introduces a checkbox to opt-in sensitive data (when unchecked).

### DIFF
--- a/src/sass/app.scss
+++ b/src/sass/app.scss
@@ -25,6 +25,7 @@
 // Components: Micro layout files. Your styles for buttons and navigation and
 // similar page components.
 //@import 'components/lists';
+@import 'components/input';
 @import 'components/spinner';
 @import 'components/links';
 @import 'components/scrollbars';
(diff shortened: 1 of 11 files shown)


lgrahl:
Will fix tests asap.

src/sass/base/_fixes.scss
```diff
@@ -0,0 +1,6 @@
+// Do not display an error spacer underneath input/textarea fields if it's empty
+.md-errors-spacer {
+    &:empty {
+        display: none;
+    }
+}
```
dbrgn: Since you're applying this to all md elements no matter where on the page, couldn't that in some cases modify the layout and/or cause layout jumps in case the error spacer is actually being used?

In doubt, just apply it to the dialog in order not to cause side effects. Otherwise, make sure that we're not relying on this anywhere. (It's well possible that it has no unintended side effects whatsoever, I'm just saying that we should double-check :slightly_smiling_face:)

lgrahl: > Since you're applying this to all md elements no matter where on the page, couldn't that in some cases modify the layout and/or cause layout jumps in case the error spacer is actually being used?

Correct. I can move it into a separate class if you want me to.

dbrgn: Yeah, maybe `md-skip-error-spacer` or something like that.

src/sass/app.scss
```diff
@@ -10,6 +10,7 @@
 @import 'base/typography';
 @import 'base/colors';
 @import 'base/cloak';
+@import 'base/fixes';
```
dbrgn: If you apply this fix using a class, make sure to move the file to the `components` directory.

PR #883, opened 2019-08-07
Fix about dialog

### LINKED ISSUE #849: About page does not show LICENSE_LINK
## Bug Description

Screenshot_2019-07-10 Threema Web

## Your Environment

* Threema Web version: Version 2.2.0-beta.3 Glärnisch
* Threema App version: 4.0-beta2 (Android)

dbrgn:
Fixes #849.

### DIFF
--- a/src/partials/messenger.ts
+++ b/src/partials/messenger.ts
@@ -44,6 +44,7 @@ import {TimeoutService} from '../services/timeout';
 import {VersionService} from '../services/version';
 import {WebClientService} from '../services/webclient';
 import {controllerModelHasMembers, isContactReceiver} from '../typeguards';
+import {Type} from '../types/helpers';
 
 // Type aliases
 import ControllerModelMode = threema.ControllerModelMode;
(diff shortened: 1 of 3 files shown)


src/partials/messenger.ts
```diff
@@ -1031,9 +1041,12 @@ class NavigationController {
     /**
      * Show dialog.
      */
-    public showDialog(name, ev) {
+    public showDialog(name, ev, controller?) {
```
lgrahl: You should be able to use default arguments here. Also, maybe add types.

```
public showDialog(name: string, ev: <no-idea>, controller: DialogController = DialogController) {
```

dbrgn: The class `DialogController` is not of type `DialogController` :wink:

lgrahl: You're missing the point. :wink:

dbrgn: Was about to give up on the `controller` type until I found the `Type` hack.

PR #889, opened 2019-08-08
Conversation updates: Log error if receiver is unknown

dbrgn:
This prevents an ugly exception in case something goes wrong.

### DIFF
--- a/src/services/webclient.ts
+++ b/src/services/webclient.ts
@@ -3165,6 +3165,11 @@ export class WebClientService {
 
         // Get receiver
         const receiver = this.receivers.getData({type: data.type, id: data.id});
+        if (!hasValue(receiver)) {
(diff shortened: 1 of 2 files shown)


src/services/webclient.ts
```diff
@@ -3108,6 +3108,11 @@ export class WebClientService {
 
         // Get receiver
         const receiver = this.receivers.getData({type: data.type, id: data.id});
+        if (!hasValue(receiver)) {
+            const log = `Received conversation update for unknown receiver (${data.type}/${data.id})`;
```
lgrahl: Would leak the receiver's id in the reported log.

dbrgn: Fixed, good catch.

dbrgn:
Ping @lgrahl

PR #894, opened 2019-08-27
Handle rejected pushes gracefully

dbrgn:
If a push was rejected with a HTTP 400 response (which can indicate an
expired push token), don't show a protocol error. Instead, offer to
forget & recreate the session.

Before this PR, a protocol error was shown.

### DIFF
--- a/src/exceptions.ts
+++ b/src/exceptions.ts
@@ -18,10 +18,11 @@
 export class TimeoutError extends Error {}
 
 export class PushError extends Error {
-    public readonly responseCode: number;
+    // HTTP response status code returned by push relay server.
+    public readonly statusCode: number;
 
-    public constructor(msg: string, responseCode: number) {
+    public constructor(msg: string, statusCode: number) {
         super(msg);
-        this.responseCode = responseCode;
+        this.statusCode = statusCode;
     }
 }
--- a/src/exceptions.ts
+++ b/src/exceptions.ts
@@ -16,3 +16,12 @@
  */
 
 export class TimeoutError extends Error {}
+
+export class PushError extends Error {
+    public readonly responseCode: number;
+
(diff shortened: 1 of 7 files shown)


src/partials/messenger.ts
```diff
@@ -165,6 +165,40 @@ export class DeviceUnreachableController extends DialogController {
...
+    }
+
+    /**
+     * Remove the stored session.
+     */
+    public forget(): void {
+        this.log.info('Forgetting stored session');
+        this.trustedKeyStore.clearTrustedKey();
+        this.cancel();
+        this.$window.location.reload();
```
dbrgn: Lazy solution to ensure that all data is gone.

src/partials/dialog.push-rejected.html
```diff
@@ -0,0 +1,25 @@
+<md-dialog aria-label="Device Unreachable">
```
lgrahl: Probably copy & paste leftover?

src/services/push.ts
```diff
@@ -124,6 +124,7 @@ export class PushSession {
      *
      * @throws TimeoutError in case the maximum amount of retries has been
      *   reached.
+     * @throws PushError if the push was rejected by the push relay server.
```
lgrahl: IIRC `Error` is not being used any more

src/exceptions.ts
```diff
@@ -16,3 +16,12 @@
  */
 
 export class TimeoutError extends Error {}
+
+export class PushError extends Error {
+    public readonly responseCode: number;
```
lgrahl: Nit: Would prefer `code` or `statusCode`.

src/services/push.ts
dbrgn: It is.

PR #904, opened 2019-09-19
CI Updates

dbrgn:
Do some UI changes according to updated CI (Corporate Identity, not Continuous Integration) guidelines.

- Hide green bar
- Show connection status in dot only when the config flag is set (enabled on staging and beta)
- Update colors of default theme
- Add work theme, depending on isWork flag in clientInfo


src/config.ts
```diff
@@ -73,5 +73,7 @@ export default {
     MSGPACK_LOG_TRACE: false,
     // Transport log level
     TRANSPORT_LOG_LEVEL: 'warn',
+    // Always show the real connection state using the dot in the logo
```
lgrahl:
```suggestion
    // Show the real connection state using the dot in the logo
```

src/partials/messenger.ts
```diff
@@ -94,22 +98,26 @@ class SendFileController extends DialogController {
...
-        'StateService', 'WebClientService',
-    ];
     private readonly $rootScope: any;
     private readonly $window: ng.IWindowService;
     private readonly stateService: StateService;
     private readonly webClientService: WebClientService;
     public retrying: boolean = false;
     public progress: number = 0;
 
+    public static readonly $inject = [
```
lgrahl: Nit-Question: Shouldn't we apply the same logic here regarding "if it doesn't fit a line, use one line per parameter"?

src/partials/messenger.ts
```diff
@@ -175,14 +183,18 @@ export class PushRejectedDialogController extends DialogController {
 
     private readonly trustedKeyStore: TrustedKeyStoreService;
 
-    public static readonly $inject = ['$mdDialog', '$window', 'TrustedKeyStore' , 'LogService'];
+    public static readonly $inject = [
```
lgrahl: ^

src/partials/messenger.ts
```diff
@@ -215,11 +225,18 @@ class SettingsController extends DialogController {
     private notificationPreview: boolean;
     private notificationSound: boolean;
 
-    constructor($mdDialog: ng.material.IDialogService,
-                $window: ng.IWindowService,
-                settingsService: SettingsService,
-                notificationService: NotificationService) {
-        super($mdDialog);
+    public static $inject = [
```
lgrahl: ^

lgrahl:
Since I haven't tested it, yet: Does this also remove the "Connection problems" text in the oragen/red bar? If not, we should do that.

dbrgn:
> IIRC, it should be fairly easy to make this persistent by using local storage. I'd recommend doing that.

To make what persistent?

dbrgn:
> To make what persistent?

From your comment, I assumed that you want to make the theme selection persistent (store the user's theme). That doesn't work in general, since the theme depends on the device / app used, so if you use a different app you'd get a different theme.

However, we can persist the theme selection if the user stores a session, together with the session keys etc. I'll implement that.

dbrgn:
I just realized that session keys are persisted during the welcome phase and the theme is only determined once the `ClientInfo` message arrives. So when we persist the keys, we don't know the theme yet. Furthermore, we cannot read the data in the keystore before the user has entered the password.

We also have the SettingsService that stores certain settings in the local storage without encryption. However, then the setting would be detached from the current session. What we could do is the following:

- Store the theme setting in the local storage without encryption
- When receiving the `ClientInfo` message, overwrite the stored theme
- When loading the page without a stored session, clear the stored theme
- When clearing a stored session, clear the stored theme
- When loading the page with a stored session, load the stored theme

Of course it could still happen that the stored theme does not match the stored session, but that would require a logic bug or manual modifications in the local storage. However, it should work in most cases.

Of course we could also just avoid the additional effort and load the theme when connecting to the session. Before that stage, the page looks almost identical (the only visible difference is the link color).

lgrahl:
> Before we receive the `ClientInfo` message, the page looks almost identical in both themes (the only visible difference is the link color).

Okay, fine be me for now. https://github.com/threema-ch/threema-web/pull/904#issuecomment-533211144 remaining then.

dbrgn:
> Does this also remove the "Connection problems" text in the oragen/red bar? If not, we should do that.

No.

img

Why should it be removed?

dbrgn:
We decided to keep the connection problems text for now and deal with it later.

PR #918, opened 2019-10-23
Handling https://threema.id links

### LINKED ISSUE #877: Handle https://threema.id/<ID> URL
We should handle `https://threema.id/<ID>` like `threema://`. Both *compose* and *add* are specified for `threema.id` links.

sillych:
Fix #877

### DIFF
--- a/src/directives/click_action.ts
+++ b/src/directives/click_action.ts
@@ -33,7 +33,7 @@ export default [
     ) {
 
         const validateThreemaId = (id: string): boolean => {
-            return id !== undefined && id !== null && /^[0-9A-Z]{8}/.test(id);
+            return id !== undefined && id !== null && /^([a-zA-Z0-9\*]{1}[a-zA-Z0-9]{7})$/.test(id);
         };
         const viewReceiver = (receiver: threema.Receiver) => {
             return function(e: Event) {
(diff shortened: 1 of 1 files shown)


src/directives/click_action.ts
```diff
@@ -33,7 +33,7 @@ export default [
     ) {
 
         const validateThreemaId = (id: string): boolean => {
-            return id !== undefined && id !== null && /^[0-9A-Z]{8}/.test(id);
+            return id !== undefined && id !== null && /^([a-zA-Z0-9\*]{1}[a-zA-Z0-9]{7})$/.test(id);
```
dbrgn: The `{1}` is optional.

```suggestion
            return id !== undefined && id !== null && /^([a-zA-Z0-9\*][a-zA-Z0-9]{7})$/.test(id);
```

src/directives/click_action.ts
```diff
@@ -104,13 +104,21 @@ export default [
...
+                                            if (matches !== null) {
+                                                const handler = getThreemaActionHandler(matches[1]);
+                                                const params = uriService.parseQueryParams(matches[2]);
+                                                if (handler !== null && params !== null) {
+                                                    node.addEventListener('click', handler(params));
+                                                }
+                                            }
+                                        } else if (link.toLowerCase().startsWith('https://threema.id/')) {
+                                            // tslint:disable-next-line:max-line-length
+                                            const matches = (/\bhttps:\/\/threema\.id\/([a-zA-Z0-9\*]{1}[a-zA-Z0-9]{7})\b/gi).exec(link);
```
dbrgn: Same here:

```suggestion
                                            const matches = (/\bhttps:\/\/threema\.id\/([a-zA-Z0-9\*][a-zA-Z0-9]{7})\b/gi).exec(link);
```

dbrgn:
Hm, when I click on a link, the add action is triggered but the page is still opened. Maybe the event needs to be cancelled manually?

sillych:
Fixed

src/directives/click_action.ts
```diff
@@ -49,6 +49,8 @@ export default [
                     return false;
                 }
 
+                e.preventDefault();
```
dbrgn: Thanks! I think previously this was not required because `threema://...` is not actually a valid URL :slightly_smiling_face:

Issue #18, opened 2017-02-15
Scrolling up sometimes fails


dbrgn:
E.g. scrolling up does not work with the "home" key if conversation is scrolled all the way to the bottom.

I also heard reports that scrolling fails with a touchpad.

Most probably has to do with the angularjs-scroll-glue library.

patrick-tee:
Confirmed: Touchpad scrolling up using a Samsung Galaxy Tab S2 jumps back to the end after half a second, no matter if the scrolling gesture is still in progress or if the finger is up again.

lgrahl:
I'm also experiencing this on FF 55 (jumps back to the bottom if not scrolling further than a certain amount).

mist:
This is really anoying, I'm using FF57 on Debian and current FF on Android and cannot see anything but the last few messages, because the site scrolls to the last message after less than a second.
Please fix this issue!

rugk:
What do you mean with desktop? The chat overview where you can select the chat/contact?

mist:
On my Desktop Computer. While on my Tablet Computer (Firefox@Android) I cannot use my cursor buttons because it hase none.

studersi:
+1

I'm running Threema Web 1.8.2 on Linux Mint 18 in Firefox 58.0.2 (64-bit). It happens when I'm scrolling with my notebook's tablet as well as with my Logitech mouse.

dbrgn:
@studersi does the problem also occur with Chromium? Maybe it is triggered by some smooth scrolling settings in Firefox...

studersi:
@dbrgn You're right, it does seem to be related to smooth scrolling to an extent.

I don't have Chromium installed but I do have Google Chrome [Version 64.0.3282.186 (Official Build) (64-bit)]. It scrolls just fine on Chrome.

In Firefox, smooth scrolling is disabled. When I enable smooth scrolling, the problem disappears **almost** completely. It's still noticeable sometimes but it just seems to stutter before scrolling as planned. Very rarely, it stays at the bottom of the conversation.

mist:
I can kind of confirm that It has to do with smooth scrolling.
Got a new machine with more power and without smooth scrolling set the problem kind of is there, although just annoying but not rendering the page useless. When I enable smooth scrolling the problem is gone.
On the Android tablet I see no way to enable smooth scrolling, so the problem persists.

dbrgn:
[...]

PR #923, opened 2019-11-10
Fix scroll glue behavior

vortex852456:
The throttleling of the scroll event causes the behavior in issue https://github.com/threema-ch/threema-web/issues/18.
But to be able to detach from bottom-glue without throtteling, updateScrollJump() must already have set showScrollJump to true when there is scrolled more than one pixel.

I guess this solution is not optimal for performance, but I have no problems in my browser. 
It's much better than living with the scroll-glue bug, though.

### DIFF
--- a/src/partials/messenger.ts
+++ b/src/partials/messenger.ts
@@ -410,7 +410,7 @@ class ConversationController {
                 $rootScope.$apply(() => {
                     this.updateScrollJump();
                 });
-            }, 100, this), supportsPassive() ? {passive: true} : false);
+            }, 0, this), supportsPassive() ? {passive: true} : false);
         }
 
         // Set receiver, conversation and type
@@ -915,12 +915,12 @@ class ConversationController {
     }
 
     /**
-     * Only show the scroll to bottom button if user scrolled more than 10px
+     * Only show the scroll to bottom button if user scrolled more than 1px
      * away from bottom.
      */
     private updateScrollJump(): void {
         const chat = this.domChatElement;
-        this.showScrollJump = chat.scrollHeight - (chat.scrollTop + chat.offsetHeight) > 10;
+        this.showScrollJump = chat.scrollHeight - (chat.scrollTop + chat.offsetHeight) > 1;
     }
 
     /**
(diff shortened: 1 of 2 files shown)


vortex852456:
Feel free to test at http://threema-web.vortex852456.de/ (I will shutdown the service after merge, though).

vortex852456:
Feel free to test at https://threema-web.vortex852456.de/
But I will shut down this demo in a few weeks.

dbrgn:
Thanks, I'll test this for a while!

dbrgn:
Seems to work nicely so far, haven't noticed big performance issues! I'd say let's promote this to `master`, we'll probably create a new beta soon.

Thanks, @vortex852456!

src/partials/messenger.ts
```diff
@@ -410,7 +410,7 @@ class ConversationController {
                 $rootScope.$apply(() => {
                     this.updateScrollJump();
                 });
-            }, 100, this), supportsPassive() ? {passive: true} : false);
+            }, 0, this), supportsPassive() ? {passive: true} : false);
```
dbrgn: Oh, before we merge this, can you completely remove the `throttle` call? It only causes unnecessary overhead when called with `0` as parameter.

vortex852456: Good point. But I had problems when I initially removed the whole throttle-call. Anyway it could have been a side effect of some other changes I did not commit (I tried different approaches for several hours and then I was just glad that it was working).

So I will try it without the throttle and report back.

Thanks for reviewing!

dbrgn: Yep, unfortunately this bug is the result of a race condition, so since removing the throttle function removes a layer of dispatching, it *could* make a difference whether it's there or not. Maybe you can test again?

The long-term solution will be a rewrite of the chat view. But a fix/workaround for now is great.

vortex852456: Yeah, I also see this only as a workaround. I guess the root of the problem is `scroll-glue="!ctrl.showScrollJump"` in `messenger.converstion.html`.  It's accessing a (for 100ms) cached value since the value of `showScrollJump` only changes with a 100ms delay (caused by the throttle). But since the `scroll-glue` has no throttle, it overwrites the scroll position before `showScrollJump` is set. 

Additionally the 10px buffer for setting `showScrollJump` to true makes it even harder to detach from the `scroll-glue`.

I successfully tested it without the throttle (see commit 7842e0a) and updated my demo.

dbrgn: Great, thank you!

Issue #197, opened 2017-04-16
Use slow KDF to derive local storage key


simmac:
While looking through the source code of the keystore service, the [pwToKey](https://github.com/threema-ch/threema-web/blob/master/src/services/keystore.ts#L78-L86) function caught my attention.

For deriving the key from the password, a SHA-512 hash is calculated and truncated to 32 bytes. So here is my first question: Why would you truncate a SHA-512 hash instead of using SHA-256 which returns exactly the 32 bytes needed? The difference between SHA-256 and SHA-512 is pretty much only the word size, so why not use the algorithm with the fitting word size (if you aren't using a slow hashing function anyways)?

Which leads me to my second question: Have you thought about using PBKDF2 for deriving the key? I would imagine that most users use relatively weak passwords for Threema Web, because they have to enter it very often (I am also guilty of this), and PBKDF2 would slow down dictionary attacks dramatically.

(Because this "issue" isn't a critical security problem per se and could benefit from a public discussion, I chose to open an issue on GitHub instead of contacting the security contact from README)

dbrgn:
> So here is my first question: Why would you truncate a SHA-512 hash instead of using SHA-256 which returns exactly the 32 bytes needed?

The reason is that [tweetnacl-js](https://github.com/dchest/tweetnacl-js) provides a SHA512 implementation, but no SHA256 :)

You're right about using a slow KDF, that would indeed be an option for the future. Note however that an attacker would first have to get access to your local storage in order to steal the hashed password, in which case you probably have bigger problems. But in any case, thanks for the suggestion, we'll think about it.

simmac:
Ah, that explains a lot for me :)

> Note however that an attacker would first have to get access to your local storage [...], in which case you probably have bigger problems.

Yes, I agree, that's why I would prefer to have the ability to save a session without a password (which would have some additional benefits: auto-connect after reloading the page or a connectivity change for example) and make the encryption password an option (just like the passphrase in the Android application).

And just for clarification:

> in order to steal the **hashed password**

The hashed password is never stored if I am not mistaken, as it acts as an encryption key for the data in the local storage, right? The threat would be someone stealing the local storage data and brute-forcing the decryption by using just the hashes (or first halves of the hashes in this case) of a rainbow table as key candidates, they would never actually have to crack the password itself.

dbrgn:
> The hashed password is never stored if I am not mistaken, as it acts as an encryption key for the data in the local storage, right?

Yes, you're right of course.

[...]

PR #929, opened 2019-11-21
Add graphical password strength indicator

dbrgn:
Show a rough indicator of password strength.

img

img

img

The checker won't prevent anyone from using a password it deems insecure. This is still the responsibility of the user.


ovalseven8:
#197 would be good complementary to this PR.

PR #934, opened 2019-12-10
More emoji fixes

### LINKED ISSUE #926: Some emoji not inserted properly
## Bug Description

Some emoji don't show up as image when inserted.

Insert `:woman_emoji:`

img

## Potential Solution

The emoji variant selector is missing in the generated filename.

dbrgn:
There were still some problems with #933. Fixes #926.


dbrgn:
Hm, a bug remains: Exclamation mark without emoji variant modifier is not converted to the proper filename by `parseEmoji`.

tools/twemoji/generate-shortname-mapping.py
```diff
@@ -14,8 +14,7 @@
             mappings.append((shortname.strip(':'), emoji['emoji']))
 
 # Custom shortcodes
-mappings.append(('slight_smile', '🙂'))
-mappings.append(('slight_frown', '🙁'))
+#mappings.append(('slight_smile', '🙂'))
```
lgrahl: :question:

dbrgn: I left it in there as documentation, in case we want to add custom shortcodes without having to upstream them first.

(The two that I removed are already upstreamed.)

dbrgn:
> Is there a documentation / procedure for what is needed to do when updating emojis?

Pray :pray:

(Basically, run `make` in the "twemoji-picker" repo in our internal gitlab, then copy over the generated assets and run the two scripts in `tools/twemoj/`. So far, something has broken on every update, so the whole process is not really stable. I hope it gets better, now that we use emojibase as data source, which is generated directly from unicode data files.)

tools/twemoji/generate-shortname-mapping.py
dbrgn: I replaced it with an actual shortcode I just found to be missing :slightly_smiling_face:

PR #938, opened 2019-12-10
Add length limits to text input fields

dbrgn:
- Limit nickname to 32 characters
- Limit first name, last name, group name, distribution list name to
  256 characters

In case multi-byte characters are used, the values will be rejected by
the apps with the error `valueTooLong`. I don't think we need to add
special handling / validation for these rare cases.

### DIFF
--- a/public/i18n/en.json
+++ b/public/i18n/en.json
@@ -249,7 +249,8 @@
             "disabledByPolicy": "Feature disabled by administrator",
             "syncFailed": "Group synchronization failed",
             "noMembers": "No members defined",
-            "alreadyLeft": "You have already left this group"
+            "alreadyLeft": "You have already left this group",
+            "valueTooLong": "A value you entered is too long"
         }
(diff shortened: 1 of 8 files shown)


public/i18n/en.json
```diff
@@ -249,7 +249,8 @@
             "disabledByPolicy": "Feature disabled by administrator",
             "syncFailed": "Group synchronization failed",
             "noMembers": "No members defined",
-            "alreadyLeft": "You have already left this group"
+            "alreadyLeft": "You have already left this group",
+            "valueTooLong": "A value you entered is too long"
```
lgrahl: In hindsight, I think this should have been just `invalidValue` because it is so generic anyway...

dbrgn: If we wanted it generic we could have used `badRequest`. But I think `valueTooLong` is fine :slightly_smiling_face:

PR #943, opened 2019-12-17
Wait briefly for a broadcast message before attempting to connect

### LINKED ISSUE #927: Don't connect if another session is active
## Bug Description

We recently made a change in the app so that a push with the same affiliation id will stop and discard any already existing Threema Web session.

Because pushes for an already active session were previously ignored, we detected this (since #369) and show a warning. However, the push is not prevented.

This means that now a second tab with the same session will quickly flash the "already connected" dialog, but will the continue connecting. The other tab will eventually lose the connection.

1. With the current `develop` version of the Android app and Threema Web e76643b746a572ba4bf77334c74a85125656e623 (now on staging), establish a stored session
2. Connect to the same session in a second tab

## Potential Solution

- Variant A: Avoid connecting or pushing if another tab is active. This probably requires an ugly hack with a timeout, because the broadcast channel is not synchronous.
- Variant B: Remove duplicate session detection. This will cause "unexplainable connection failures" with older app versions though. Additionally, when a session is "taken over", the old session is not immediately terminated but might try to reconnect, potentially causing a situation where both sessions are continuously reconnecting.

lgrahl:
Resolves #927

### DIFF
--- a/src/partials/welcome.ts
+++ b/src/partials/welcome.ts
@@ -273,18 +273,22 @@ class WelcomeController {
         // Set up the broadcast channel that checks whether we're already connected in another tab
         this.setupBroadcastChannel(this.webClientService.salty.keyStore.publicKeyHex, 0)
             .then((result) => {
-                switch (result) {
-                    case 'already_open':
-                        this.log.warn('Session already connected in another tab or window');
-                        break;
-                    case 'no_answer':
-                        this.start();
-                        break;
-                }
+                this.$scope.$apply(() => {
+                    switch (result) {
+                        case 'already_open':
+                            this.log.warn('Session already connected in another tab or window');
+                            break;
+                        case 'no_answer':
+                            this.start();
+                            break;
+                    }
+                });
             })
             .catch((error) => {
-                this.log.warn('Unable to set up broadcast channel:', error);
-                this.start();
+                this.$scope.$apply(() => {
+                    this.log.warn('Unable to set up broadcast channel:', error);
+                    this.start();
+                });
             });
     }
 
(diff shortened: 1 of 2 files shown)


src/partials/welcome.ts
```diff
@@ -301,24 +317,43 @@ class WelcomeController {
...
-        this.reconnect(keyStore, decrypted);
+        this.setupBroadcastChannel(keyStore.publicKeyHex, WelcomeController.BROADCAST_DELAY)
+            .then((result) => {
+                this.$scope.$apply(() => {
+                    switch (result) {
+                        case 'already_open':
+                            this.log.warn('Session already connected in another tab or window');
+                            break;
+                        case 'no_answer':
+                            this.log.debug('No broadcast received indicating that a session is already open');
```
dbrgn: We log this line here, but not in the "start" case. Do we need it at all, or was it mainly for debugging?

src/partials/welcome.ts
```diff
@@ -262,18 +263,33 @@ class WelcomeController {
...
         this.$scope.$watch(() => this.password, () => {
             const payload = this.webClientService.buildQrCodePayload(this.password.length > 0);
             this.qrCode = this.buildQrCode(payload);
             this.passwordStrength = scorePassword(this.password);
         });
 
-        // Start webclient
-        this.start();
+        // Set up the broadcast channel that checks whether we're already connected in another tab
+        this.setupBroadcastChannel(this.webClientService.salty.keyStore.publicKeyHex, 0)
```
dbrgn: Why a timeout of 0? Won't this immediately resolve to `no_answer`?

tsconfig.json
```diff
@@ -1,8 +1,9 @@
 {
     "compilerOptions": {
-        "target": "ES2017",
-        "module": "esNext",
-        "moduleResolution": "node",
+        "lib": ["DOM", "ESNext"],
+        "target": "ESNext",
```
dbrgn: Just out of curiosity: Why was this change required?

src/partials/welcome.ts
lgrahl: Yup. Should be valid for the *scan* case.

src/partials/welcome.ts
lgrahl: Because in the `scan` case, there's no delay. That's the main difference.

tsconfig.json
lgrahl: TS didn't like `Promise.finally`.

src/partials/welcome.ts
dbrgn: Ah, I see, in this case we set up the broadcast channel so it can answer requests from other tabs, but we don't expect a valid answer ourselves (since this is a new connection) :slightly_smiling_face:

PR #953, opened 2020-01-07
Ignore case when translating shortcodes

dbrgn:
Low hanging :cherries:, for people who get stuck on the shift key after entering `:`.


dbrgn:
Hm, I removed the configurability. On second thought, that's a case of speculative generality and YAGNI.

PR #893, opened 2019-08-21
Fix race conditions

lgrahl:
Fix a bunch of race conditions.

PR #954, opened 2020-01-08
Revert PR #893

dbrgn:
The PR potentially introduced connectivity problems. Revert it to be
able to compare the behavior.


lgrahl:
Looks like the demon portal was only partially closed.

PR #956, opened 2020-01-09
Value too long error

dbrgn:
Some workarounds to alleviate message length problems reported through support.

### DIFF
--- a/src/directives/compose_area.ts
+++ b/src/directives/compose_area.ts
@@ -202,7 +202,14 @@ export default [
                         // Check for max length
                         const byteLength = (new TextEncoder().encode(text)).length;
                         if (byteLength > scope.maxTextLength) {
-                            const pieces: string[] = stringService.byteChunk(text, scope.maxTextLength, 50);
+                            // Messages that use the entire available message size become a problem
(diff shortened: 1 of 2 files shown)


src/directives/compose_area.ts
```diff
@@ -202,7 +202,7 @@ export default [
                         // Check for max length
                         const byteLength = (new TextEncoder().encode(text)).length;
                         if (byteLength > scope.maxTextLength) {
-                            const pieces: string[] = stringService.byteChunk(text, scope.maxTextLength, 50);
+                            const pieces: string[] = stringService.byteChunk(text, scope.maxTextLength * 0.85, 50);
```
lgrahl: Eh?

dbrgn: Did you read the commit message?

dbrgn: (A `Math.floor` call is missing though...)

lgrahl: Umm, ok, but since it's a workaround, perhaps add a comment?

dbrgn: Ok, fair point.

PR #958, opened 2020-01-13
Do not show protocol error alert when DC closes

dbrgn:
...but do not remove `failSession`, as we did in #893.

### DIFF
--- a/src/services/webclient.ts
+++ b/src/services/webclient.ts
@@ -1095,8 +1095,8 @@ export class WebClientService {
                 });
             };
             dc.onclose = () => {
-                this.arpLog.error(`Data channel ${dc.label} closed prematurely`);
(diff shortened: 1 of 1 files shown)


lgrahl:
Rather port the whole `onclose` event handler from #893 as is because if we enter the demon portal (i.e. the WebSocket create -> close loop), the data channel never closes because it has not been created. So this **should** be safe. But we should monitor it regardless.

dbrgn:
@lgrahl including or excluding `failSession()`? IIRC you remarked that removing `failSession` could be a problem as well.

lgrahl:
Anything that triggers the loop, yeah. This could be a trigger for the demon portal but it's less risky than the rest and I think it's worth a shot.

PR #962, opened 2020-01-23
Update compose-area to v0.3.5

dbrgn:
Just a maintenance release.


dbrgn:
That there are no changes in how it works, the library simply updated its dependencies: https://github.com/threema-ch/compose-area/blob/master/CHANGELOG.md#v035-2020-01-21

PR #968, opened 2020-02-18
Upgrade dependencies

dbrgn:
Split into multiple commits so that potential regressions can be found more easily by bisecting.

I checked the changelogs of most relevant packages to avoid bad surprises.


dbrgn:
Upgrade party :smile:

PR #969, opened 2020-02-20
Work around event handling changes in Chrome

lgrahl:
Chrome now doesn't fire ICE state change events when the peer connection is being
closed by the application (which is probably spec-compliant).
But Chrome now also (always?) fires the 'error' event on the data
channel when it's being closed implicitly by the peer connection (wtf).

### DIFF
--- a/src/services/peerconnection.ts
+++ b/src/services/peerconnection.ts
@@ -46,6 +46,7 @@ export class PeerConnectionHelper {
 
     // Handed over signalling channel
     private sdc: UnboundedFlowControlledDataChannel | null = null;
+    private closed: boolean = false;
 
     // Calculated connection state
     public connectionState: TaskConnectionState = TaskConnectionState.New;
@@ -150,7 +151,12 @@ export class PeerConnectionHelper {
                             // situations which certainly would lead to ugly race conditions.
                             this.connectionFailedTimer = null;
                             this.log.debug('ICE connection considered failed');
-                            this.pc.close();
+
+                            // Close and raise disconnected event
+                            // Note: We need to simulate a 'closed' event here due to browser inconsistencies
+                            this.closed = true;
+                            this.close();
+                            this.setConnectionState(TaskConnectionState.Disconnected);
                         }, PeerConnectionHelper.CONNECTION_FAILED_TIMEOUT_MS, true, 'connectionFailedTimer');
                         break;
                     case 'connected':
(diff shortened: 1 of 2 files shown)


src/services/peerconnection.ts
```diff
@@ -150,7 +151,12 @@ export class PeerConnectionHelper {
                             // situations which certainly would lead to ugly race conditions.
                             this.connectionFailedTimer = null;
                             this.log.debug('ICE connection considered failed');
-                            this.pc.close();
+
+                            // Close and raise disconnected event
+                            // Note: We need to simulate a 'closed' event here due to browser inconsistencies
+                            this.closed = true;
```
dbrgn: The PeerConnectionHelper is never reused, right? (Asking since `closed` is never reset.)

src/services/peerconnection.ts
```diff
@@ -250,8 +261,12 @@ export class PeerConnectionHelper {
 
             // Rebind close event
             dc.onclose = () => {
-                this.log.info(`Signalling data channel closed`);
-                link.closed();
+                if (this.closed) {
```
dbrgn: You're using `this` here and `self` above... Should probably be `self` in both cases for explicitness (although `this` should work as well).

src/services/peerconnection.ts
```diff
@@ -285,6 +300,13 @@ export class PeerConnectionHelper {
      * Unbind all event handler and abruptly close the peer connection.
      */
     public close(): void {
+        // Cancel connection failed timer
+        if (this.connectionFailedTimer !== null) {
+            this.timeoutService.cancel(this.connectionFailedTimer);
+            this.connectionFailedTimer = null;
```
dbrgn: When you call `close()` from [here](https://github.com/threema-ch/threema-web/pull/969/files#diff-f0ddf3d90ae1f0bfd272dc1a4e352307R158), the timer will always have been set to null (without cancelling it?). But maybe we should still keep the code above, just to make sure that the timeout is cancelled?

src/services/peerconnection.ts
```diff
@@ -250,8 +261,12 @@ export class PeerConnectionHelper {
 
             // Rebind close event
             dc.onclose = () => {
-                this.log.info(`Signalling data channel closed`);
-                link.closed();
+                if (this.closed) {
+                    self.log.debug('Ignoring signalling data channel closed');
```
lgrahl:
```suggestion
                    this.log.debug('Ignoring signalling data channel closed');
```

src/services/peerconnection.ts
lgrahl: Yup, never reused.

src/services/peerconnection.ts
lgrahl: Yup, since it can be called from the outside.

PR #972, opened 2020-03-12
Add service worker for PWA

chrisly-bear:
Related issue: #407

Added minimal service worker javascript code required to have Chrome (on Desktop) recognize the app as a PWA and make it installable. This allows users to install the app on their desktops and have it behave like a native application.


public/manifest.webmanifest
```diff
@@ -4,7 +4,7 @@
   "icons": [
     {
       "src": "img/favicon/favicon.ico?v=[[VERSION]]",
-      "sizes": "64x64",
+      "sizes": "48x48",
```
lgrahl: Can you elaborate on this change?

public/js/service-worker-registration.js
```diff
@@ -0,0 +1,11 @@
+// register service worker
```
lgrahl: Instead of having a separate file, can you integrate the registration into https://github.com/threema-ch/threema-web/blob/168c60310fd6b3331b3c1fe18c50db4a49994b78/src/bootstrap.ts#L26 (before that line)?

public/js/service-worker-registration.js
```diff
@@ -0,0 +1,11 @@
...
+
+if ('serviceWorker' in navigator) {
+    navigator.serviceWorker.register('/service-worker.js')
+        .then(function (registration) {
+            // console.log('Service worker registration successful, scope is:', registration.scope);
```
lgrahl: Having this enabled in the logs in production is fine from my perspective.

```suggestion
            console.info('Service worker registration successful, scope is:', registration.scope);
```

It would be even better to have it in its own service since it could then use the `Logger` service but it should be fine for now.

public/js/service-worker-registration.js
```diff
@@ -0,0 +1,11 @@
...
+        })
+        .catch(function (error) {
+            // console.log('Service worker registration failed, error:', error);
```
lgrahl:
```suggestion
            console.warn('Service worker registration failed, error:', error);
```

public/manifest.webmanifest
chrisly-bear: The favicon `public/img/favicon/favicon.ico` has resolution 48x48. Chrome DevTools report an error if we declare it as 64x64 resolution. It is not a crucial change (the PWA is still installable by Chrome), but I just noticed the error and thought I'd fix it.

resolution_error

public/js/service-worker-registration.js
chrisly-bear: Did that. Thanks for the hint (I wasn't sure where to put the code).

public/js/service-worker-registration.js
chrisly-bear: I went with your suggestion and enabled the `console.info` statement.

public/js/service-worker-registration.js
chrisly-bear: I enabled the `console.warn` statement.

public/manifest.webmanifest
lgrahl: Makes sense. Good catch. :slightly_smiling_face:

PR #979, opened 2020-04-01
Design updates

dbrgn:
- New logo
- Use Lab Grotesque font for officially hosted versions of Threema Web


dbrgn:
> Resolving this at runtime delays loading of the stylesheet file.

True, but with browser caching this should not be a big issue in practice. Even if you just visited the Threema website before, the font and its CSS will already be cached.

PR #980, opened 2020-04-02
Proper handling of invalid contacts

dbrgn:
Contact state can be invalid, for example if the identity has been revoked.

In the detail view, it now looks like this:

invalid

Sending messages to invalid contacts is handled the same way as messages to blocked contacts (sending will be denied with a toast).


src/partials/messenger.receiver/contact.html
```diff
@@ -10,7 +10,8 @@
 				<dt>Threema ID</dt>
 				<dd>
 					<span class="complex-values">
-						<span>{{ctrl.receiver.id}}</span>
+						<span>{{ ctrl.receiver.id }}</span>
+						<span ng-if="ctrl.receiver.state !== 'ACTIVE'">({{ ctrl.receiver.state }})</span>
```
lgrahl, line 14: The raw untranslated state is now exposed in the UI. Is this acceptable?

dbrgn: Originally that was my intent, yes, but I changed my mind (since we can pull most translation strings from the app).

dbrgn:
Now with translation and additional "revoked"-terminology:

inactive

revoked

PR #981, opened 2020-04-02
Catch and log unhandled exceptions

dbrgn:
This way exceptions also land in the troubleshooting log.

Here's an example of a log with the following cases covered:

- Uncaught exception
- Unhandled promise rejection (with `Error` as reject reason)
- Unhandled promise rejection (with string as reject reason)
- Uncaught exception in a callback (scheduled with `setTimeout`)

2020-04-02-153357_1823x761_scrot

Some of these cases can be handled by registering error handlers on the window object, others are caught by AngularJS and need to be processed using a `$exceptionHandler` factory.

### DIFF
--- a/src/app.ts
+++ b/src/app.ts
@@ -24,6 +24,7 @@ import config from './config';
 import './controllers';
 import './directives';
 import './filters';
+import {hasValue} from './helpers';
 import './partials/messenger';
 import './partials/welcome';
 import './services';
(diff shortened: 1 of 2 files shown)


dbrgn:
I'll handle the failing unit tests a bit later today. They happen because the logger itself now logs something on initialization, which confuses the tests.

src/app.ts
```diff
@@ -184,4 +185,21 @@ angular.module('3ema', [
     }]);
 }])
 
+.factory('$exceptionHandler', ['LogService', function(logService: LogService) {
+    const logger = logService.getLogger('UncaughtException');
+    return function myExceptionHandler(exception: Error, cause) {
+        const data: any[] = [];
+        const message = exception.message.includes(exception.name)
+            ? exception.message
+            : `${exception.name}: ${exception.message}`;
```
dbrgn, line 194: Aaaaah I can't do that because it's possible to throw non-`Error` types :roll_eyes:

PR #992, opened 2020-05-30
Fix typing indicator on focus loss

### LINKED ISSUE #924: Typing notification/"message" not cancelled when tab is switched
## Bug Description

When you do not stop typing, but immediately switch to another tab, Threema Web doe snot notice you've stopped typing, i.e. the user user thinks you are still typing while you are actually browsing the web in another tab.

1. Start typing something in conversation.
2. Switch tabs. (Ctrl+Tab likely, IIRC)

**What happens:** other chat partner continued to see "user is typing…" even though I stopped.

**What should happen:** stop typing "message" should be sent

3. After I switched to the tab back again, after one second or less it seems to have detected that I stopped typing and the status changed.

(only noticed one time, did not try to reproduce myself)

## Potential Solution

detect tab change and react

## Your Environment

* Threema Web version: 2.2.1-gh
* Threema App version: 4.21
* Browser name and version: Firefox 70
* Computer operating system and version: Linux
* Smartphone operating system and version: Android 9

dennisheer:
Sends stopTyping message if window loses focus. Fixes #924.

Running the tests as instructed in the [contribution document](https://github.com/threema-ch/threema-web/blob/master/CONTRIBUTING.md) introduced no new failed tests.

### DIFF
--- a/src/directives/compose_area.ts
+++ b/src/directives/compose_area.ts
@@ -721,24 +721,25 @@ export default [
 
                 updateView();
 
-                const stopTypingOnBlur = () => stopTyping();
+                // Callbacks to unsubscribe listeners in $destroy
+                const unsubscribeListeners = [];
 
+                // Send "stop typing" message when switching tab or window
+                const stopTypingOnBlur = () => stopTyping();
                 window.addEventListener('blur', stopTypingOnBlur);
+                unsubscribeListeners.push(() => window.removeEventListener('blur', stopTypingOnBlur));
 
                 // Listen to broadcasts
-                const unsubscribeListeners = [];
                 unsubscribeListeners.push($rootScope.$on('onQuoted', (event: ng.IAngularEvent, args: any) => {
-                        composeArea.focus();
-                    }),
-                    () => window.removeEventListener('blur', stopTypingOnBlur)
-                );
+                    composeArea.focus();
+                }));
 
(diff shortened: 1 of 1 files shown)


src/directives/compose_area.ts
```diff
@@ -735,6 +735,9 @@ export default [
                     });
                     stopTyping();
                 });
+
+                // When losing browser or tab focus, send stopTyping message
+                window.onblur = () => stopTyping();
```
dbrgn, line 740: Overwriting `onblur` is not an ideal solution because it might overwrite other event handlers that want to listen to the `blur` event. Furthermore, the event is never deregistered, so the `blur` event will trigger even if no chat is open.

The better solution would be to use the `unsubscribeListeners` hack used above:

- Store the callback in a local variable: `const stopTypingOnBlur = () => stopTyping()`
- To register the event listener, use `window.addEventListener('blur', stopTypingOnBlur)`
- To ensure that the listener is deregistered properly, push `() => window.removeEventListener('blur', stopTypingOnBlur)` into `unsubscribeListeners`

This is all untested, so please check whether it works as intended :slightly_smiling_face:

dennisheer: Resolved in 50060caa21295ca4e7f79195d4dd2837a8973fd9.

PR #1014, opened 2020-12-30
Fix logic for determining whether a point in time was yesterday

### LINKED ISSUE #1013: logic for determining whether a point in time was yesterday is flawed
## Bug Description

The logic for determining whether a point in time was yesterday https://github.com/threema-ch/threema-web/blob/39c046ffb5e9d06b1e7cad346df99d1f9766aee6/src/filters.ts#L385 doesn't work correctly around leap seconds and DST changes.

For days longer than 24 h (i.e., leap seconds and "fall back" out of DST), this doesn't matter, as the lines before it https://github.com/threema-ch/threema-web/blob/39c046ffb5e9d06b1e7cad346df99d1f9766aee6/src/filters.ts#L381-L383 will already correctly identify any point in time of the same date as being on the same day, even if that point in time is more than 24 h before `now`.

For days shorter than 24 h (i.e., "spring forward" into DST), this however meant they will incorrectly be "skipped" when looking from a point in time on the next day but less than 24 h after the beginning of said shorter day (i.e. between 0:00 AM and 1:00 AM on the day _after_ the shorter day) back to points in time within that shorter day. Here's a test to demonstrate the problem: https://github.com/threema-ch/threema-web/blob/a3de76459d1f37b2c03a48d98e0faf1c4e729c5b/tests/filters.js#L199-L213 All 3 assertions fail when run in a timezone that switched for standard time to DST on 2020-03-29, such as `Europe/Zurich`.

This problem was found in the source code, not by observing the application behavior. But I guess it could be reproduced like so (untested):

1. Be in a timezone that observes DST
2. On the day DST begins, send a message.
3. On the next day, between 0:00 AM (midnight) and 1:00 AM, look at that message

### Expected

The day of the message is displayed as "yesterday".

### Observed (speculative, as I didn't test this)

The day of the message is displayed its date.

## Potential Solution

Use approach from [this Stack Overflow answer](https://stackoverflow.com/a/5511376/674064) instead: https://github.com/threema-ch/threema-web/blob/a3de76459d1f37b2c03a48d98e0faf1c4e729c5b/src/filters.ts#L385-L386

## Your Environment

* Threema Web version: Source code on `master` (39c046ffb5e9d06b1e7cad346df99d1f9766aee6)
* Browser name and version: Firefox 84.0 (64-bit)
* Computer operating system and version: [NixOS](https://nixos.org/) unstable:
  - system: `"x86_64-linux"`
  - host os: `Linux 5.4.85, NixOS, 21.03pre259798.84917aa00bf (Okapi)`
  - multi-user?: `yes`
  - sandbox: `yes`
  - version: `nix-env (Nix) 2.3.10`
  - channels(root): `"nixos-21.03pre259798.84917aa00bf, nixos-unstable-21.03pre257780.e9158eca70a"`
  - channels(das-g): `""`
  - nixpkgs: `/nix/var/nix/profiles/per-user/root/channels/nixos`

das-g:
Fixes #1013

The previous logic https://github.com/threema-ch/threema-web/blob/39c046ffb5e9d06b1e7cad346df99d1f9766aee6/src/filters.ts#L385 didn't work correctly around leap seconds and DST changes.

For days longer than 24 h (i.e., leap seconds and "fall back" out of DST), this didn't matter, as the lines before it https://github.com/threema-ch/threema-web/blob/39c046ffb5e9d06b1e7cad346df99d1f9766aee6/src/filters.ts#L381-L383 would already have correctly identified any point in time of the same date as being on the same day, even if that point in time was more than 24 h before `now`.

For days shorter than 24 h (i.e., "spring forward" into DST), this however meant they could incorrectly be "skipped" when looking from a point in time on the next day but less than 24 h after the beginning of said shorter day (i.e. between 0:00 AM and 1:00 AM on the day _after_ the shorter day) back to points in time within that shorter day, as demonstrated by the added test: https://github.com/threema-ch/threema-web/blob/a3de76459d1f37b2c03a48d98e0faf1c4e729c5b/tests/filters.js#L199-L213 All 3 assertions fail without the change in `src/filters.ts`, when run in a timezone that switched for standard time to DST on 2020-03-29, such as `Europe/Zurich`. For the test to portably demonstrate the problem, the timezone would have to be mocked, for which jasmine doesn't seem to have native support. I've tried to use the approach from [How to mock your browser’s timezone with Jasmine and MomentJS](https://philippe.bourgau.net/how-to-mock-your-browsers-timezone-with-jasmine-and-momentjs/) in the branch [`mock-timezone`](https://github.com/das-g/threema-web/tree/mock-timezone) ([diff](https://github.com/das-g/threema-web/compare/fix-yesterday...mock-timezone)), but am unsure how to make `moment` available within `tests/filters.js`.

Despite looking funky, the new logic https://github.com/threema-ch/threema-web/blob/a3de76459d1f37b2c03a48d98e0faf1c4e729c5b/src/filters.ts#L385-L386 which takes an approach from [this Stack Overflow answer](https://stackoverflow.com/a/5511376/674064) works also across month and year boundaries, as demonstrated by another test: https://github.com/threema-ch/threema-web/blob/a3de76459d1f37b2c03a48d98e0faf1c4e729c5b/tests/filters.js#L215-L225 According to [a comment on said Stack Overflow answer](https://stackoverflow.com/questions/5511323/calculate-the-date-yesterday-in-javascript#comment34446144_5511376), it does so across multiple browsers.

### To do

Even without the fix, the test to demonstrate this issue passes in timezones that don't switch to DST on on 2020-03-29. Making the test timezone-aware probably requires either to use `moment-timezone` or more elaborate manual mocking. For either I'd need further guidance and/or someone else would have to implement them.

### DIFF
--- a/src/filters.ts
+++ b/src/filters.ts
@@ -382,7 +382,8 @@ angular.module('3ema.filters', [])
             return formatTime(date);
         }
 
-        const yesterday = new Date(now.getTime() - 1000 * 60 * 60 * 24);
+        const yesterday = new Date(now);
+        yesterday.setDate(now.getDate() - 1);
         if (!forceFull && isSameDay(date, yesterday)) {
             return $translate.instant('date.YESTERDAY') + ', ' + formatTime(date);
         }
--- a/tests/filters.js
+++ b/tests/filters.js
@@ -196,6 +196,18 @@ describe('Filters', function() {
             ]);
         });
 
+        it('shows "yesterday" for yesterday also around DST change', () => {
+            // This test relies on being run in timezone Europe/Zurich
+            // FIXME: Mock timezone, so that test is portable.
+            const beforeDST = new Date(2020, 2, 29, 0, 42);
+            const duringDST = new Date(2020, 2, 30, 0, 23);
+            jasmine.clock().install();
+            jasmine.clock().mockDate(duringDST);
+            this.testPatterns([
+                [beforeDST.getTime() / 1000, 'date.YESTERDAY, 00:42'],
+            ]);
+        });
+
         it('shows full datetime for other days', () => {
             jasmine.clock().install();
             jasmine.clock().mockDate(new Date(2018, 9, 9, 20, 42));
--- a/tests/filters.js
+++ b/tests/filters.js
@@ -196,22 +196,6 @@ describe('Filters', function() {
             ]);
         });
 
-        it('shows "yesterday" for yesterday also around DST change', () => {
-            // This test relies on being run in timezone Europe/Zurich
-            // FIXME: Mock timezone, so that test is portable.
-            const beforeDST1 = new Date(2020, 2, 29, 0, 42);
-            const beforeDST2 = new Date(2020, 2, 29, 12, 42);
-            const beforeDST3 = new Date(2020, 2, 29, 23, 59);
-            const duringDST = new Date(2020, 2, 30, 0, 23);
-            jasmine.clock().install();
-            jasmine.clock().mockDate(duringDST);
-            this.testPatterns([
-                [beforeDST1.getTime() / 1000, 'date.YESTERDAY, 00:42'],
-                [beforeDST2.getTime() / 1000, 'date.YESTERDAY, 12:42'],
-                [beforeDST3.getTime() / 1000, 'date.YESTERDAY, 23:59'],
-            ]);
-        });
-
         it('works across month and year boundaries', () => {
             const in20thCentury1 = new Date(1999, 11, 30, 23, 59);
             const in20thCentury2 = new Date(1999, 11, 31, 00, 00);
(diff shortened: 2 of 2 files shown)


dbrgn:
Thanks for the fix and the detailed explanations!

I think the fix itself should be good to merge. Regarding the timezone, since Threema Web will be sunset this year, it probably isn't worth the added dependencies and complexity to mock the timezone. What do you think about removing the DST test and just keeping the other one?

das-g:
> What do you think about removing the DST test and just keeping the other one?

Fine with me. In fact, that's one the options I suggested in https://github.com/threema-ch/threema-web/issues/1013#issuecomment-753486954. I'll adapt the pull request accordingly.

dbrgn (approved):
Thanks!

PR #1019, opened 2021-01-07
Use Ctrl+Enter to submit instead of Shift+Enter

dbrgn:
This is a more common key combination.

@jaremforster could you test this on macOS, to verify that the Cmd key is properly recognized both in Safari and a third party browser (e.g. Chrome / Firefox)?


lgrahl (approved):
No objections. Works on my machine in Chromium and Firefox.

Issue #92, opened 2017-02-23
add X to clear search bar


SimJoSt:
## Expected Behavior
Clear search input by click on a clear search-bar button.

## Current Behavior
Cannot find clear search-bar button when done with a search, clear it by hand (marking everything, delete).

## Possible Solution
Add an X button at the end of the search bar, to clear previously entered input.

## Steps to Reproduce (for bugs)
Click on the search icon, input a string, do your thing, be finished.
Want to revert to empty search bar to have all my conversations back in the list, click on clear search-bar button, done.

*I am aware, that clicking on the seach icon again hides and clears the search-bar. But not everybody (myself included, was smart enough to find it at the first try). Also I like the possibility to temporarily deactivate the search filter and activate it again when I need it. To achieve that, the behaviour of the second click on the search icon need to be altered to not clear the input anymore.*

Buddabread:
It would be nice if the Search-bar would be automatically cleared after a contact was found and clicked.
After selection of a contact I would like to have the normal view with recent messages.
In my opinion this would be more intuitive.

samuelT2:
Additionally, it would be nice if pressing ESC is binded to the same function as hiding/showing the search icon.

samuelT2:
Hi @dbrgn - I'd like to start working on this issue, fine for you?

dbrgn:
Hi @samuelT2. Yes, as long as the implementation doesn't get overly complicated, we'll happily review PRs!

samuelT2:
Hi @dbrgn - Got a solution ready which covers all above points, but does not support Firefox (because of the use of [::-webkit-search-cancel-button](https://developer.mozilla.org/en-US/docs/Web/CSS/::-webkit-search-cancel-button)). You want me to create the pull request and review the change or shall I invest some time to implement a proper custom solution?

dbrgn:
It would be worthwhile to at least investigate whether an easy alternative solution would be possible.

If not, as long as the input field can be cleared with ESC, I guess it's OK.

samuelT2:
Created a pull request - bringing the X to non-webkit-supported browsers needs more than a dozen lines of CSS. ESC though, works perfectly fine now on all browsers. Think this should do it.

PR #1026, opened 2021-01-19
Allow easily clearing search box

samuelT2:
Fix related to "add X to clear search bar", issue #92 

- Clear search bar with ESC or X (X only works in Chrome)
- Hide search bar when conversation or contact is clicked

### DIFF
--- a/src/partials/messenger.navigation.html
+++ b/src/partials/messenger.navigation.html
@@ -64,7 +64,7 @@
         </md-button>
     </div>
     <div class="search" searchbox searchbox-focus="ctrl.searchVisible">
-        <input type="text" ng-model="ctrl.searchText" translate translate-attr-placeholder="messenger.SEARCH" aria-label="Search" translate-attr-aria-label="messenger.SEARCH">
+        <input type="search" ng-model="ctrl.searchText" translate translate-attr-placeholder="messenger.SEARCH" aria-label="Search" translate-attr-aria-label="messenger.SEARCH" ng-keyup="ctrl.clearSearch($event)">
     </div>
 </div>
 
@@ -81,6 +81,7 @@
             ng-if="ctrl.isVisible(conversation)">
 
             <div class="conversation"
+                  ng-click="ctrl.toggleSearchOff()"
                   ng-class="{'unread': conversation.unreadCount > 0,
                             'starred': conversation.isStarred,
                             'active': ctrl.isActive(conversation)}">
@@ -126,7 +127,8 @@
 <div id="navigation-contacts" class="tab-content" ng-if="ctrl.activeTab == 'contacts'" in-view-container>
     <p class="empty" ng-if="ctrl.contacts().length === 0" translate>messenger.NO_CONTACTS_FOUND</p>
     <ul ng-class="{'hide-inactive': ctrl.hideInactiveContacts()}">
-        <li ng-repeat="contact in ctrl.contacts() | filter:ctrl.searchContact"
+        <li ng-click="ctrl.toggleSearchOff()"
+            ng-repeat="contact in ctrl.contacts() | filter:ctrl.searchContact"
             ui-sref="messenger.home.conversation({ type: 'contact', id: contact.id, initParams: null })"
             class="contact"
             ng-class="{'inactive': contact.state == 'INACTIVE'}">
--- a/src/partials/messenger.navigation.html
+++ b/src/partials/messenger.navigation.html
@@ -64,7 +64,7 @@
         </md-button>
     </div>
     <div class="search" searchbox searchbox-focus="ctrl.searchVisible">
-        <input type="search" ng-model="ctrl.searchText" translate translate-attr-placeholder="messenger.SEARCH" aria-label="Search" translate-attr-aria-label="messenger.SEARCH" ng-keyup="ctrl.clearSearch($event)">
+        <input type="search" ng-model="ctrl.searchText" translate translate-attr-placeholder="messenger.SEARCH" aria-label="Search" translate-attr-aria-label="messenger.SEARCH" ng-keydown="ctrl.clearSearch($event)">
     </div>
 </div>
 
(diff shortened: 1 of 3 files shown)


samuelT2:
CLA is signed, not sure why it's still showing as not signed.

samuelT2:
CLA shows as not signed because my e-mail address is not linked to my commit (now it is, but I cannot modify existing commits).

dbrgn:
> CLA shows as not signed because my e-mail address is not linked to my commit (now it is, but I cannot modify existing commits).

You can easily fix this:

    git config user.email desired-email@example.com
    git commit --amend --reset-author
    git push --force

samuelT2:
Very cool, did solve it, thanks!

src/partials/messenger.ts
```diff
@@ -1267,6 +1267,21 @@ class NavigationController {
         this.searchVisible = !this.searchVisible;
     }
 
+    /**
+     * Toggle search bar off.
+     */
+    public toggleSearchOff(): void {
+        if (this.searchVisible) this.searchVisible = false;
```
dbrgn, line 1274: Nitpick: Can you add explicit curly braces? We prefer that for explicitness.

```suggestion
        if (this.searchVisible) {
            this.searchVisible = false;
        }
```

src/partials/messenger.ts
```diff
@@ -1267,6 +1267,21 @@ class NavigationController {
...
+    }
+
+    /**
+     * Clear search when pressing ESC.
+     */
+    public clearSearch(ev): void {
+        if (ev.keyCode === 27) this.searchText = '';
```
dbrgn, line 1281: The `keyCode` field is deprecated and should not be used anymore: https://developer.mozilla.org/en-US/docs/Web/API/KeyboardEvent/keyCode Instead, use `ev.key === 'Escape'`.

```suggestion
        if (ev.key === 'Escape') {
            this.searchText = '';
        }
```

dbrgn (changes requested):
Thanks! Some notes below.

samuelT2:
Requested changes are done, anything else needed from my side? Sorry that I ask, my very first pull request...

dbrgn:
Thanks for applying the fixes!

> Requested changes are done, anything else needed from my side? Sorry that I ask, my very first pull request...

No, you did everything right. We're currently juggling with different tasks and priorities (especially now that a lot new users joined), so it may take a few days until we can review a submission!

I tested your changes, they seem to work nicely in Chromium! I like that the search box disappears when clicking on a conversation, that makes sense. However, in Firefox the ESC key does not seem to trigger the keyUp event 🤔 Maybe you can test as well? It's possible that Firefox does not consider all key types for the keyUp event.

I also had another thought: Right now ESC clears the input field if it contains a search. Maybe it could also hide the search box if it's already empty? This way, a search could be aborted by pressing ESC twice.

samuelT2:
Sure - no hurry!

> I tested your changes, they seem to work nicely in Chromium! I like that the search box disappears when clicking on a conversation, that makes sense. However, in Firefox the ESC key does not seem to trigger the keyUp event 🤔 Maybe you can test as well? It's possible that Firefox does not consider all key types for the keyUp event.

With which Firefox version are you testing currently? That ESC event works fine me with, it clears the search field in Chrome and Firefox 84.0.2 (64-Bit).

> I also had another thought: Right now ESC clears the input field if it contains a search. Maybe it could also hide the search box if it's already empty? This way, a search could be aborted by pressing ESC twice.

This one is nice, and not much of work. Will do it and push it again.

dbrgn:
> With which Firefox version are you testing currently? That ESC event works fine me with, it clears the search field in Chrome and Firefox 84.0.2 (64-Bit).

Hm, that's strange, I'm also on 84.0.2 and for me the KeyUp listener is never even called. I'll wait for your update, and will then test again to find out what the reason is.

src/partials/messenger.navigation.html
```diff
@@ -64,7 +64,7 @@
         </md-button>
     </div>
     <div class="search" searchbox searchbox-focus="ctrl.searchVisible">
-        <input type="text" ng-model="ctrl.searchText" translate translate-attr-placeholder="messenger.SEARCH" aria-label="Search" translate-attr-aria-label="messenger.SEARCH">
+        <input type="search" ng-model="ctrl.searchText" translate translate-attr-placeholder="messenger.SEARCH" aria-label="Search" translate-attr-aria-label="messenger.SEARCH" ng-keyup="ctrl.clearSearch($event)">
```
threema-danilo, line 67: There's one last remaining problem: With the change in https://github.com/threema-ch/threema-web/pull/1026/commits/837762a4b2efa403787de6f32bc58d5dcc5acdb6, pressing ESC in Chrome with a filled search box will immediately close it.

I believe this is because search fields are automatically cleared by Chrome when pressing ESC, so at the time of the keyup event, it's already empty.

The solution to this is to change the line above from ng-keyup to ng-keydown. This will also result in a _slightly_ faster reaction.

threema-danilo (changes requested):
One last change request, then this is ready to merge.

threema-danilo (approved):
Thanks @samuelT2!

threema-danilo:
Deployed to https://web-beta.threema.ch/, testing welcome. If no major bugs in that release are discovered in the next 2-3 days, a public release will be made.

threema-danilo:
Unfortunately even small features can have undesired side effects...

It seems that when searching through a lot of conversations (I have over 230 conversations), after clicking on a conversation, the search clears almost instantly, but it sometimes takes multiple seconds until the conversation has opened.

This has a lot to do with the inefficient architecture of AngularJS, where every UI update can trigger a lot of processing code, which in turn may cause yet another UI update.

For now, we might need to disable the clearing/hiding of the search when clicking on a search result.

samuelT2:
Ouch, well - didn't test it with 230+ conversations :). Shall I revert that change or will you do it?

threema-danilo:
See https://github.com/threema-ch/threema-web/pull/1042

PR #1027, opened 2021-01-19
Add white margin around QR code

dbrgn:
image

This allows scanning the QR code even if a browser extensions overrides the CSS background color (e.g. some "dark theme" or "high contrast" browser extensions), a problem that was frequently reported to our support.

(Because the angular-qrcode lib seems unmaintained, I had to vendor it. Modifications were applied in a separate commit.)


lgrahl (approved):
LGTM.

Nit: In web terminology, _margin_ should perhaps preferably be called _padding_ or _border_.

dbrgn:
> Nit: In web terminology, margin should perhaps preferably be called padding or border.

In the box model, margin is outside the boundary and padding is inside the boundary. In other words, it's a margin around the QR code shape and a padding of the canvas. Or an invisible border, as the border color and backgrond color are identical 🤷🏻‍♂️

I just reused the terminology of the angular-qrcode patch 😉

Issue #1009, opened 2020-12-24
Minimal UI option


rasenplanscher:
I'd like to use ThreemaWeb in Rambox, and the header, footer, and padding are taking up too much space. Please use a focussed design, at least for small viewports. WhatsApp Web does that well: There's only the list of contacts and the message area.

I don't need a header telling me this is Threema, I know that, so it can just go. Also, I frankly don't care about the footer items 99% of the time, so they can be safely hidden in the **⋮**-menu.

dbrgn:
Please note that Threema Web is in [maintenance mode](https://github.com/threema-ch/threema-web/pull/996) so no new features will be added.

(If a PR comes in with a change that's very simple and is unlikely to introduce any new bugs, it will be considered for merging though!)

jwanner83:
I'm interested in implementing this feature. Is it okay if I give it a shot? :)

jwanner83:
I thought it would make sense to have a new setting called "User Interface / Benutzeroberfläche" where you can change from "default" to "minimal". This setting is then stored in the localStorage.

### Setting
<img width="830" alt="Bildschirmfoto 2021-01-28 um 10 40 18" src="https://user-images.githubusercontent.com/38656104/106119220-3a1c9c00-6155-11eb-9f3d-506a62a45bd1.png">

It then adds a class to the body which makes it possible to style certain elements the way we want (like to remove the header / footer ...)

### Default User Interface:
<img width="1473" alt="grafik" src="https://user-images.githubusercontent.com/38656104/106120261-743a6d80-6156-11eb-99d0-c5a05bb36c14.png">

### Minimal User Interface:
<img width="1475" alt="grafik" src="https://user-images.githubusercontent.com/38656104/106120288-7e5c6c00-6156-11eb-9ea8-bd1dc34132a3.png">

To still be able to see the version of the application, I added a button which opens a version dialog. This button is only visible if the minimal theme is active.

#### Button
<img width="265" alt="Bildschirmfoto 2021-01-28 um 10 51 00" src="https://user-images.githubusercontent.com/38656104/106120488-b95e9f80-6156-11eb-9ef6-b2ac70e8b716.png">

#### Dialog
<img width="528" alt="Bildschirmfoto 2021-01-28 um 10 51 28" src="https://user-images.githubusercontent.com/38656104/106120539-c8dde880-6156-11eb-9faa-05c6b7a2c353.png">


[...]

PR #1029, opened 2021-01-28
#1009 add minimal user interface

jwanner83:
This adds the option of a minimal user interface. The changes I've made are described in the following comment. https://github.com/threema-ch/threema-web/issues/1009#issuecomment-768938478

### DIFF
--- a/index.html
+++ b/index.html
@@ -67,7 +67,7 @@
     <link rel="stylesheet" href="css/app.css?v=[[VERSION]]">
 </head>
 
-<body ng-controller="ThemeController as themeCtrl" class="{{ [ctrl.statusClass, themeCtrl.userInterfaceClass].join(' ') }}" ng-class="{expanded: ctrl.expandStatusBar}" md-theme="{{ themeCtrl.theme }}">
+<body ng-controller="ThemeController as themeCtrl" class="{{ [ctrl.statusClass, themeCtrl.userInterfaceClass].join(' ') }}" ng-class="{expanded: ctrl.expandStatusBar, messenger: ctrl.showsMessenger()}" md-theme="{{ themeCtrl.theme }}">
     <img src="img/bg.jpg?v=[[VERSION]]" aria-label="Background image: Blurred photo of a mountain" id="background-image" draggable="false">
 
     <noscript class="compat compat-nojs">
@@ -94,7 +94,7 @@
         </div>
     </div>
 
-    <div id="main-wrapper" ng-cloak ng-class="{wide: ctrl.wide()}">
+    <div id="main-wrapper" ng-cloak ng-class="{wide: ctrl.showsMessenger()}">
         <header>
             <h1 id="title" aria-label="Threema Web Logo">
                 <span id="logo" ng-include src="'img/logo.svg?v=[[VERSION]]'"></span>
--- a/index.html
+++ b/index.html
@@ -67,7 +67,7 @@
     <link rel="stylesheet" href="css/app.css?v=[[VERSION]]">
 </head>
 
-<body ng-controller="ThemeController as themeCtrl" class="{{ [ctrl.statusClass, themeCtrl.userInterfaceClass].join(' ') }}" ng-class="{expanded: ctrl.expandStatusBar, messenger: ctrl.showsMessenger()}" md-theme="{{ themeCtrl.theme }}">
+<body ng-controller="ThemeController as themeCtrl" class="{{ [ctrl.statusClass, themeCtrl.userInterfaceClass].join(' ') }}" ng-class="{expanded: ctrl.expandStatusBar, messenger: ctrl.showMessenger()}" md-theme="{{ themeCtrl.theme }}">
     <img src="img/bg.jpg?v=[[VERSION]]" aria-label="Background image: Blurred photo of a mountain" id="background-image" draggable="false">
 
     <noscript class="compat compat-nojs">
@@ -94,7 +94,7 @@
         </div>
     </div>
 
-    <div id="main-wrapper" ng-cloak ng-class="{wide: ctrl.showsMessenger()}">
+    <div id="main-wrapper" ng-cloak ng-class="{wide: ctrl.showMessenger()}">
         <header>
             <h1 id="title" aria-label="Threema Web Logo">
                 <span id="logo" ng-include src="'img/logo.svg?v=[[VERSION]]'"></span>
--- a/index.html
+++ b/index.html
@@ -67,7 +67,7 @@
     <link rel="stylesheet" href="css/app.css?v=[[VERSION]]">
 </head>
 
-<body ng-controller="ThemeController as themeCtrl" class="{{ ctrl.statusClass }}" ng-class="{expanded: ctrl.expandStatusBar}" md-theme="{{ themeCtrl.theme }}">
+<body ng-controller="ThemeController as themeCtrl" class="{{ [ctrl.statusClass, themeCtrl.userInterfaceClass].join(' ') }}" ng-class="{expanded: ctrl.expandStatusBar}" md-theme="{{ themeCtrl.theme }}">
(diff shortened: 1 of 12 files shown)


src/sass/ui/_minimal.scss
```diff
@@ -0,0 +1,26 @@
+.user-interface {
+    &--minimal {
+        &.messenger {
+            header,
+            footer {
+                display: none;
```
jwanner83, line 6: The reason I did it with `display: none;` instead of `ng-hide` or `ng-show` is because these other two options do exactly the same: they just add a class and hide it with `display: none;`

aalolexx:
Very nice improvement! I like it.

pbuw:
Looks awesome!

src/sass/ui/_minimal.scss
```diff
@@ -0,0 +1,26 @@
+.user-interface {
+    &--minimal {
```
threema-danilo, line 2: Any reason for picking a double-dash as separator? At first I mis-interpreted this as a [custom CSS property](https://developer.mozilla.org/en-US/docs/Web/CSS/--*).

public/i18n/de.json
```diff
@@ -313,6 +314,11 @@
             "COMPOSE_AREA": "Eingabefeld",
             "SUBMIT_KEY_ENTER": "'Eingabe'  zum Senden einer Nachricht. 'Umschalt+Eingabe' zum Hinzufügen einer neuen Zeile.",
             "SUBMIT_KEY_CTRL_ENTER": "'Strg+Eingabe' zum Senden einer Nachricht. 'Eingabe' zum Hinzufügen einer neuen Zeile."
+        },
+        "ui": {
+           "UI": "Benutzeroberfläche",
```
threema-danilo, line 319: The indentation is wrong for this item.

src/services/settings.ts
```diff
@@ -53,6 +53,38 @@ class ComposeAreaSettings {
...
+    public setUserInterface(userInterface: string | threema.UserInterface): void {
+        this.settingsService.storeUntrustedKeyValuePair('userInterface', this.parseUserInterface(userInterface).toString());
+    }
+
+    private parseUserInterface(userInterface: any): threema.UserInterface {
+        try {
+            userInterface = parseInt(userInterface, 10);
+        } catch {
+            // Ignored
+        }
```
threema-danilo, line 81: Wouldn't it be better to just treat the user interface variant as a string, instead of converting it to an integer (and then back to a string for storage)? This would also make the template radio button values more readable. (As far as I know, TypeScript const enums can contain strings as well.)

index.html
```diff
@@ -67,7 +67,7 @@
     <link rel="stylesheet" href="css/app.css?v=[[VERSION]]">
 </head>
 
-<body ng-controller="ThemeController as themeCtrl" class="{{ ctrl.statusClass }}" ng-class="{expanded: ctrl.expandStatusBar}" md-theme="{{ themeCtrl.theme }}">
+<body ng-controller="ThemeController as themeCtrl" class="{{ [ctrl.statusClass, themeCtrl.userInterfaceClass].join(' ') }}" ng-class="{expanded: ctrl.expandStatusBar}" md-theme="{{ themeCtrl.theme }}">
```
threema-danilo: Would `class="{{ ctrl.statusClass }} {{ themeCtrl.userInterfaceClass }}"` work as well? That would be a bit easier to read.

src/controllers/theme.ts
```diff
@@ -30,18 +31,34 @@ export class ThemeController {
...
+        this.userInterfaceClass = ThemeController.getUserInterfaceClass(settingsService.userInterface.getUserInterface())
+
         // Listen to theme changes
         themeService.evtThemeChange.attach((newTheme: threema.Theme) => {
             this.log.debug(`Updating theme: ${this.theme} -> ${newTheme}`);
             $scope.$apply(() => this.theme = newTheme);
         });
     }
+
+    private static getUserInterfaceClass (userInterface: threema.UserInterface): string {
```
threema-danilo, line 62: Nitpick: The space between the function name and the parameters should be removed.

src/sass/ui/_minimal.scss
threema-danilo: Yeah, I think it's good to keep theme overrides entirely in CSS.

src/controllers/status.ts
```diff
@@ -305,7 +305,7 @@ export class StatusController {
         }, startTimeout);
     }
 
-    public wide(): boolean {
+    public showsMessenger(): boolean {
```
threema-danilo, line 308: Can you rename this to `showMessenger()`? I think it reads better 🙂

threema-danilo (changes requested):
Thanks for the PR, the full-screen layout looks good! Also, thanks for creating small, focussed commits, makes it much easier to review 🙂

I left a few comments.

src/sass/ui/_minimal.scss
jwanner83: Yeah thats my bad. I am used to work with CSS BEM (http://getbem.com/introduction/) and there it is required to do such things with double dashes or / and double underlines. 

But as I see, this project doesn't use this standard so I will change it to a single dash :)

src/services/settings.ts
jwanner83: I agree, this would be much better :) I changed it. I did this initialy with 0 and 1 because the other settings do it like this as well. But directly with the string is much better :)

jwanner83:
I implemented your changes @threema-danilo. Thank you for the review 💯

src/services/settings.ts
lgrahl: @dbrgn Just an FYI here. The reason why I chose an integer for newer settings is this: Local storage persists across versions of Threema Web and thus works similarly to a protocol. It's easy to change the name of a mapped item like an integer. But a string ties you to that string. If it's misspelled, it stays misspelled. Same if the semantics change (even if just slightly) or are extended/specialised. One can work around this mishap in the application code by mapping it using an `enum`, but a string enum in TypeScript is IMO unnecessarily clunky compared to `type Type = 'foo' | 'bar'`.

But, granted, the whole thing was hacked in by me very quickly, so it looks ugly in the radio buttons. It doesn't need to be that way, though.

src/sass/ui/_minimal.scss
threema-danilo: No worries, I simply wasn't aware of BEM :)

src/partials/messenger.ts
```diff
@@ -1291,6 +1330,12 @@ class NavigationController {
         return this.notificationService.getDndModeSimplified(conversation);
     }
 
+    /**
+     * Return true if the minimal user interface is selected.
+     */
+    private isMinimalUserInterface(userInterface: UserInterface): boolean {
```
threema-danilo, line 1336: Since this is the only case where the type alias for `UserInterface` is being used, I'd stick with the fully qualified name:

```suggestion
    private isMinimalUserInterface(userInterface: threema.UserInterface): boolean {
```

Then you can remove the alias (`import UserInterface = threema.UserInterface`) from line 52.

threema-danilo (changes requested):
Thanks! One last issue (plus a nitpick), then it's ready to merge!

src/services/settings.ts
```diff
@@ -53,6 +53,25 @@ class ComposeAreaSettings {
...
 
+class UserInterfaceSettings {
+    private readonly settingsService: SettingsService;
+
+    constructor(settingsService: SettingsService) {
+        this.settingsService = settingsService;
+    }
+
+    public getUserInterface(): threema.UserInterface {
+        return this.settingsService.retrieveUntrustedKeyValuePair('userInterface', false) as threema.UserInterface;
```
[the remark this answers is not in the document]
jwanner83: 👍  Thanks for seeing this. I've didn't test it enough :(

threema-danilo (approved):
Looks great now, thanks.

threema-danilo:
Deployed to https://web-beta.threema.ch/, testing welcome. If no major bugs in that release are discovered in the next 2-3 days, a public release will be made.

PR #1035, opened 2021-02-09
Remove feature mask checks for files and groups

threema-danilo:
Apps have already removed these checks as well.

This should get rid of the infamous "X cannot receive files" bug that happened every once in a while.


lgrahl (approved):
Looks good on first glance, haven't tested it. Annoying we never found the real culprit.

PR #1053, opened 2021-04-23
Ensure that update/activeConversation triggers a wakeup push

threema-danilo:
Otherwise iOS devices are not woken up with the `activeConversation` message.


threema-danilo:
@lgrahl right now the "send wakeup push" and "retransmit" flags are combined. This just means that the message needs to be acked by the iOS app and doesn't have any other significant side effects, right?

lgrahl:
Yes.

PR #1055, opened 2021-05-03
CI: Use node 16

threema-danilo:


threema-danilo:
On old, circleci/node does not support Node 16 yet.

threema-danilo:
Actually, node-sass is not yet compatible with node 16. Let's revert to 14.

PR #1078, opened 2021-08-25
Correctly handle paste items when pasting images

bluec0re:
The paste handler does not correctly check for item types.

This results in pasting images to not work if they are also available as {kind: 'string', type: 'image/xpm'} in the clipboard (for example as its the case with flameshot).

### DIFF
--- a/src/directives/compose_area.ts
+++ b/src/directives/compose_area.ts
@@ -442,9 +442,9 @@ export default [
                     let fileIdx: number | null = null;
                     let textIdx: number | null = null;
                     for (let i = 0; i < items.length; i++) {
-                        if (items[i].type.indexOf('image/') !== -1 || items[i].type === 'application/x-moz-file') {
+                        if (items[i].kind === 'file' && (items[i].type.indexOf('image/') !== -1 || items[i].type === 'application/x-moz-file')) {
                             fileIdx = i;
-                        } else if (items[i].type === 'text/plain') {
+                        } else if (items[i].kind === 'string' && items[i].type === 'text/plain') {
                             textIdx = i;
                         }
                     }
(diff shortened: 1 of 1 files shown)


threema-danilo:
Thanks! How can I reproduce this issue? When I use flameshot (v0.10.1) I only get a single DataTransferItem of kind `file`.

threema-danilo:
Can you run this in your JS console, and post the output generated if you paste an image from flameshot?

    document.getElementById('composeDiv').addEventListener('paste', e => { const items = e.clipboardData.items; for (i=0;i<items.length;i++) { console.log(`item ${i}: ${items[i].kind} => ${items[i].type}`) }})

What version of flamshot do you use? (Output of `flameshot --version`)

bluec0re:
Sure:

```
item 0: string => BITMAP
item 1: string => MULTIPLE
item 2: string => PIXMAP
item 3: string => SAVE_TARGETS
item 4: string => TARGETS
item 5: string => TIMESTAMP
item 6: string => application/x-qt-image
item 7: string => image/bmp
item 8: string => image/cur
item 9: string => image/ico
item 10: string => image/jpeg
item 11: string => image/jpg
item 12: string => image/pbm
item 13: string => image/pgm
item 14: file => image/png
item 15: string => image/ppm
item 16: string => image/xbm
item 17: string => image/xpm
```

The problem for me is that the JS code continues to override the index variable as long as it finds an image mime type. My first approach was to collect all indices which start with `image/` and try all until one succeeds, but I then noticed that the API you're using ([getAsFile](https://developer.mozilla.org/en-US/docs/Web/API/DataTransferItem/getAsFile)) only supports `file`s anyways.

### Flameshot version
```
Flameshot v0.10.1
Compiled with Qt 5.15.2
```

### Additional info
WM: i3
OS: Debian 11 based
Browser: Chrome 93

threema-danilo:
Oh wow, I wonder why on your system you get so many  entries in the clipboard (and I wonder why binary formats are transferred as a string). On my system (Arch Linux with dwm, identical Flameshot version) I only get one.

With your list of `DataTransferItem`s, your fix makes sense, thank you!

threema-danilo (approved):

PR #1091, opened 2021-11-08
Switch to a proper password-based key derivation function (scrypt)

### LINKED ISSUE #197: Use slow KDF to derive local storage key

lgrahl:
Existing key storage will be automatically migrated on first use.

Resolves #197


lgrahl:
@threema-danilo FYI, the tests implicitly run scrypt now since they always go through a migration step.

src/services/keystore.ts
```diff
@@ -68,71 +79,77 @@ export class TrustedKeyStoreService {
...
         pushTokenType: threema.PushTokenType | null,
         password: string,
         isAutoSession: boolean,
-    ): void {
+    ): Promise<void> {
+        // Generate a random salt of 32 bytes
+        const salt = nacl.randomBytes(32);
+
+        // Determine work factor (N) to take ~500ms but use at least a work factor of 2^16.
+        // Round the work factor to the nearest power of two.
```
lgrahl, line 108:
```suggestion
        // Round the work factor to the nearest power of two.
```

It's not the nearest and behaves more like floor.

src/services/keystore.ts
```diff
@@ -155,61 +172,162 @@ export class TrustedKeyStoreService {
...
+        const encrypted = hexToU8a(parts[1]);
+        const decrypted = nacl.secretbox.open(
+            encrypted,
+            nonce,
+            this.deprecatedPwToKey((new TextEncoder()).encode(password)),
+        );
+        if (!decrypted) {
+            throw new Error('Unable to decrypt deprecated storage');
+        }
+        const token = this.decodePushToken((decrypted as Uint8Array).slice(96));
```
threema-danilo, line 237: We should probably add a null check here to avoid an NPE (if decoding failed).

src/services/keystore.ts
```diff
@@ -155,61 +172,162 @@ export class TrustedKeyStoreService {
...
         }
 
-        const nonce = hexToU8a(parts[0]);
-        const encrypted = hexToU8a(parts[1]);
-        const decrypted = nacl.secretbox.open(encrypted, nonce, this.pwToKey(password));
-        if (!decrypted) {
+        // Parse unencrypted storage data
+        let raw: any;
+        try {
+            raw = JSON.parse(this.storage.getItem(this.STORAGE_KEY));
```
threema-danilo, line 268: Here we're re-reading the data from storage even if no migration took place.

Instead, you could re-read only if the migration happened, and assign the value to `storedValue`.

threema-danilo (changes requested):
Nice, thanks! 👍🏻

src/services/keystore.ts
```diff
@@ -155,61 +172,162 @@ export class TrustedKeyStoreService {
...
+                case threema.PushTokenType.Apns:
+                    encodedString = threema.PushTokenPrefix.Apns + ':' + token;
+                    break;
+                case threema.PushTokenType.Hms:
+                    encodedString = threema.PushTokenPrefix.Hms + ':' + token;
+                    break;
+            }
+        }
+        return (encodedString == null)
+            ? new Uint8Array(0)
```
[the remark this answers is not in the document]
lgrahl: Why do you think it's unreachable? The arguments are `token: string | null, type: threema.PushTokenType | null`

src/services/keystore.ts
lgrahl: I don't understand?

threema-danilo: `decodePushToken` may return null if decoding fails (for whatever reason). If that's the case, `token.token` and `token.type` below will result in a null pointer exception / type error.

lgrahl: Okay. NPE confused me. ;) There's no exception, it would just be undefined (which is even worse).

threema-danilo: What I meant is that four lines below, `token.token` and `token.type` are accessed. And `(null).token` does result in an exception 🙂

threema-danilo (approved):

lgrahl:
Squashed. @threema-danilo feel free to hit the merge button.

PR #1092, opened 2021-11-11
Show the complete public key

Pythonix:
Instead of showing only the first 32 chars of the public key, the complete public key should be displayed.


lgrahl:
Thanks for the PR but that's a hash of the public key and not the public key itself. Furthermore, the description would need updating. I think @threema-danilo planned something regarding that.

Pythonix:
You are right, of course! Then I will close the PR when changes are planned anyway.

threema-danilo:
Yeah, we already planned to change the display of fingerprint / public key, but we're still working out the final format. I'll provide a PR in the coming days.

Thanks anyways, @Pythonix!

PR #1100, opened 2021-11-25
Reset unread count only when necessary

threema-jarem:
iOS devices are regularly disconnected, but this is normal behavior. We therefore only want to reset the unread count if the device the connection is closed for good.

### DIFF
--- a/src/services/state.ts
+++ b/src/services/state.ts
@@ -266,7 +266,7 @@ export class StateService {
     /**
      * Reset all states.
      */
-    public reset(connectionBuildupState: threema.ConnectionBuildupState = 'new', shouldResetUnreadCount : boolean = false): void {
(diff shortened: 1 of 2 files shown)


src/services/state.ts
```diff
@@ -276,6 +276,10 @@ export class StateService {
         this.state = GlobalConnectionState.Error;
         this.connectionBuildupState = connectionBuildupState;
         this.progress = 0;
-        this.unreadCount = 0;
+
+        // iOS devices are regularly disconnected, but this is normal behavior.
```
threema-danilo, line 280: Can you move this comment to the webclient service, where the `shouldResetUnreadCount` value is determined? Here it's a bit wrong, because the state service doesn't know anything about the task used.

threema-danilo (changes requested):

threema-danilo (approved):

Issue #1118, opened 2022-03-18
Firefox: PDFs are opened in same tab


threema-danilo:
## Bug Description

In current versions of Firefox, when clicking on a PDF file, it is not downloaded, but instead opened in the same tab. This affects both current and old version of Threema Web, I suspect this was caused by a behavioral change in newer Firefox versions.

* Threema Web version: 2.3.19
* Browser name and version: Firefox 98.0.1

lgrahl:
Happens on a lot of pages, so I'm not even sure we should do anything about it.

lgrahl:
Found this so far: https://bugzilla.mozilla.org/show_bug.cgi?id=1759835

threema-danilo:
A potential workaround is downloading PDF files using the media type `application/octet-stream`, one of our support staff said that this seems to help. It's a ugly hack for a problem in Firefox, but if it works it might be a valid workaround.

I'll take a closer look this afternoon. We're using a method that should *always* result in a download, so I suspect that Firefox is doing something wrong here. If I can confirm that, I'll comment on the Bugzilla ticket.

lgrahl:
Found another longer discussion we might want to chime in: https://bugzilla.mozilla.org/show_bug.cgi?id=1756980#c5 (sigh, blaming bad UX while not even knowing the use case, like SPA with state in memory... come on)

threema-danilo:
Ugh, so they _intentionally_ ignore the `download` attribute and `Content-Disposition: attachment`? https://bugzilla.mozilla.org/show_bug.cgi?id=1756980#c7

lgrahl:
It may be worth investigating whether using [modern DOM APIs](https://github.com/eligrey/FileSaver.js#saving-file) results in the same issue. I don't think we need to use `file-saver` - our browser requirements are high enough.

threema-danilo:
> We're using `file-saver` which uses `target=_blank`, so this would be the most fitting bug I guess: https://bugzilla.mozilla.org/show_bug.cgi?id=1759916

file-saver uses a lot of different approaches depending on the browser: https://github.com/eligrey/FileSaver.js/blob/master/src/FileSaver.js In our case (Firefox) an approach with `a.href = URL.createObjectURL(blob)` followed by `click(a)` is used.

Fun fact from bugzilla 1756980:

> This also happens on WhatsApp web now. PDF files are opened in the same tab, although pressing the back button does re-logins the users. But I think this shows that this behavior is not expected.

[...]

PR #1120, opened 2022-03-23
Add workaround for Firefox non-standard PDF download behavior

threema-danilo:
See https://github.com/threema-ch/threema-web/issues/1118 for more details.

With this fix, the PDF still opens immediately instead of just downloading, but in a new tab.

@lgrahl do you see any major blocker here? (Yes, I know it's ugly.)

### DIFF
--- a/src/helpers.ts
+++ b/src/helpers.ts
@@ -430,3 +430,18 @@ export function replaceWhitespace(text: string): string {
         .replace(/ /g, '&nbsp;')
         .replace(/\t/, '&nbsp;&nbsp;');
 }
+
+/**
+ * Work around nonstandard Firefox behavior when downloading a PDF by changing
+ * a PDF mimetype to application/octet-stream.
+ *
+ * See https://github.com/threema-ch/threema-web/issues/1118
+ */
+export function firefoxWorkaroundPdfDownload(mimetype: string): string {
+    const uagent = window.navigator.userAgent.toLowerCase();
+    const isFirefox = /mozilla/.test(uagent) && /firefox/.test(uagent); // Ugh
+    if (isFirefox && mimetype === 'application/pdf') {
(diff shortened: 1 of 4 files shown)


src/helpers.ts
```diff
@@ -430,3 +430,18 @@ export function replaceWhitespace(text: string): string {
...
+ *
+ * See https://github.com/threema-ch/threema-web/issues/1118
+ */
+export function firefoxWorkaroundPdfDownload(mimetype: string): string {
+    const uagent = window.navigator.userAgent.toLowerCase();
+    const isFirefox = /mozilla/.test(uagent) && /firefox/.test(uagent); // Ugh
+    if (isFirefox && mimetype === 'application/pdf') {
+        return 'application/octet-stream';
+    }
+    return mimetype;
```
lgrahl, line 446: Why don't you use the existing BrowserInfo class?

threema-danilo: I assume you mean `BrowserService`? Because it's an angular service with dependencies on other services, and I didn't want to inject it everywhere (and pass it through to the helper). Additionally, we only need to detect Firefox, the `BrowserService` does more than that.

But the Firefox detection line is taken from the `BrowserService`.

lgrahl: Ok, I don't have a strong opinion on this, just wondering.

lgrahl (approved):
The only other thing I wonder is if we want to do version detection, too. Otherwise, LGTM.

src/helpers.ts
threema-danilo: `BrowserInfo` is just the data holder returned by `BrowserService` 🙂

threema-danilo:
Thanks! For now I'd try to keep things simple and omit version detection.

PR #1124, opened 2022-04-24
Improve initial instruction when opening the webpage

### LINKED ISSUE #1112: Improve initial instruction on web.threema.ch
When navigating to web.threema.ch, the instruction how to proceed is very short.

Currently the message is:
`Scan this QR code with your Threema app`
But it is not obvious, **where** in the app you do scanning.

I suggest something like:
`Scan this QR code with your Threema app.
To do this, start the app and tap on menu (icon "...") (Android) or Settings (icon "gear wheel") on iOS, then select "Threema Web" and follow the instructions there.`

This is a small change requiring only adaption of (translated) text.
The translation label is PLEASE_SCAN in
https://github.com/threema-ch/threema-web/blob/1656f800cd9db3bda8ced7991607deba1ddab1a4/public/i18n/en.json#L8
Maybe this is adapted with this translation label, maybe a 2nd label introduced (like PLEASE_SCAN_INSTR).

SimonHeimberg:
Fixes #1112

Having icons for "menu" and "Settings" would be nice. But this is a good step.

### DIFF
--- a/src/partials/welcome.html
+++ b/src/partials/welcome.html
@@ -6,7 +6,9 @@
                 <p><strong translate>common.NOTE</strong> <span translate>welcome.SAFARI</span></p>
             </div>
 
-            <p class="instructions" translate>welcome.PLEASE_SCAN</p>
+            <p class="instructions" translate-attr="{'title:': 'welcome.PLEASE_SCAN_INSTR'}">
+                <span translate>welcome.PLEASE_SCAN</span> <i class="material-icons">info</i>
+            </p>
             <qrcode version="{{ ctrl.qrCode.version }}"
                     id="connecting-code"
                     error-correction-level="{{ ctrl.qrCode.errorCorrectionLevel }}"
--- a/public/i18n/en.json
+++ b/public/i18n/en.json
@@ -6,6 +6,7 @@
         "ABORT": "Abort",
         "CONTINUE_ANYWAY": "Continue anyway",
         "PLEASE_SCAN": "Scan this QR code with your Threema app",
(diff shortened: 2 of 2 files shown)


src/partials/welcome.html
```diff
@@ -6,7 +6,9 @@
                 <p><strong translate>common.NOTE</strong> <span translate>welcome.SAFARI</span></p>
             </div>
 
-            <p class="instructions" translate>welcome.PLEASE_SCAN</p>
+            <p class="instructions" translate-attr="{'title:': 'welcome.PLEASE_SCAN_INSTR'}">
```
threema-danilo, line 9: Stray colon in the title attribute name.

threema-danilo (changes requested):
This doesn't actually work yet, did you test it? Besides the stray colon sign in the attribute name, the information icon is very large, not centered, and a bit distracting.

screenshot-20220425-102006

Maybe something like this would be better? (Font size is `.8em`)

SimonHeimberg:
> This doesn't actually work yet, did you test it? Besides the stray colon sign in the attribute name, the information icon is very large, not centered, and a bit distracting.
> 

Sorry, no, I did not test it. I do not have the setup to do it. I have forgotten to mention this. The only thing I tested was to put the icon in the code (in the debug tools of the browser).

> 
> Maybe something like this would be better? (Font size is `.8em`)
> 
> screenshot-20220425-102754

I like your suggestion. It is short and clear. Will you implement it?

threema-danilo:
Replaced by #1126.

PR #1139, opened 2022-06-09
Clear the password field in case of incorrect password entry

Pythonix:
Currently, after the notification that the password entry was incorrect, the password must be deleted manually. This PR will automatically delete the password.

### DIFF
--- a/src/partials/welcome.ts
+++ b/src/partials/welcome.ts
@@ -572,6 +572,7 @@ class WelcomeController {
             parent: angular.element(document.body),
             clickOutsideToClose: true,
             fullscreen: true,
+            onRemoving: () => { this.clearPassword(); },
         });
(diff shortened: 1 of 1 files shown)


threema-danilo:
Thanks! That makes sense.

However, it would feel more natural if the password is cleared when the error dialog is closed. Could you change that? In `showDecryptionFailed`, you can add an `onRemoving` callback to the `.show()`-arguments:

```ts
this.$mdDialog.show({
    ...
    onRemoving: () => { this.clearPassword(); },
});
```

(Untested)

Pythonix:
Thanks, I adopted your suggestion and tested it. Works as intended.

threema-danilo:
Thanks! Looks good now 🙂

PR #1175, opened 2023-06-24
Add support for Threema Gatway push service

threema-lenny:
These pushes are relayed via Threema Gateway. Primarily for Threema Libre users and users who use Threema Push in the app who currently suffer from the degraded user experience.

Requires the deployment of https://github.com/threema-ch/push-relay/pull/52 and an update of the Android app.


threema-danilo (approved):
Untested, but looks good!

PR #1201, opened 2024-04-02
Fix compose area scroll bug

### LINKED ISSUE #1200: Compose area with multiple lines cannot be scrolled fully to the top
## Bug Description

When 15 lines are in the compose area, in some browsers (Chromium), by scrolling all the way to the top, the top few lines are invisible.

1. Use Chromium
2. Type "a"
3. Type 14 lines with "b" (use shift+enter to add newline)
4. Scroll all the way to the top, "a" is not visible

image

## Potential Solution

Add `max-height: 100%` to div

## Your Environment

* Threema Web version: 2.5.3
* Browser name and version: Chromium 123
* Computer operating system and version: Arch Linux

threema-danilo:
Fixes #1200.

### DIFF
--- a/src/sass/sections/_compose_area.scss
+++ b/src/sass/sections/_compose_area.scss
@@ -33,6 +33,7 @@ compose-area {
                         cursor: text;
                         padding: $main-padding;
                         min-height: 22px;
+                        max-height: 100%;
                         line-height: 20px;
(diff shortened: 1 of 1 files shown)


threema-gian:
Thanks, the fix works as intended. Merging now!

threema-gian (approved):

PR #1205, opened 2024-04-03
Fix latest message change detection

threema-danilo:


threema-danilo:
Note: Before merging, update the ARP docs.

PR #1220, opened 2024-10-14
Fix beta link in work flavor and open in new tab

threema-gian:
Correctly links to the work download page, and opens the beta link in a new tab.


threema-danilo (approved):
LGTM!

PR #1222, opened 2024-12-17
Minimal support for emoji reactions

threema-lenny:
Only shows reactions in their respective buckets. Still limits reactions from Threema Web to ack/dec.


threema-lenny:
@threema-gian thanks! I just shoved another commit in here (sorry). Can you give the Safari enabler a quick smoke test (should work on Mac OS but not on iOS)?

threema-gian (approved):
LGTM, thanks!

PR #1273, opened 2026-06-08
Add `id` to session restore password form

FineFindus:
Some password managers (such as Bitwarden) allow autofilling of custom fields, such as the session restore password, using the `id` of the input field to distinguish it from other password fields. Unfortunately, the session restore password field, does not have an explicit `id` set, instead using an autogenerated one (either e.g. `input_0` or `input_1`), which can cause the custom autofilling to break.
This adds an explicit `id` to the session restore password field, to allowing password managers to distinguish it from the user password field.

Ref: https://bitwarden.com/help/auto-fill-custom-fields/#using-linked-custom-fields


threema-danilo (approved):
LGTM, thanks!

